PR #6, opened 2021-03-11
updated popper.js, added 'hide' modifier to Tooltip & Dropdown

dsv-sergey:
- [Breaking Change]: In Popper 2 this is now an array of objects (modifiers), instead of an object where each property was the modifier name. [More information.](https://popper.js.org/docs/v2/migration-guide/)


uui-components/src/overlays/Dropdown.tsx
```diff
@@ -41,7 +40,7 @@ export interface DropdownProps extends Partial<IEditable<boolean>> {
     closeOnMouseLeave?: 'toggler' | 'boundary' | false;
 
     portalTarget?: HTMLElement;
-    boundaryElement?: PopperJS.Boundary | Element;
+    boundaryElement?: Boundary;
```
AlekseyManetov, line 43: In Popper 2 'Element' can't be passed?

uui-components/src/overlays/Dropdown.tsx
```diff
@@ -183,17 +180,39 @@ export class Dropdown extends React.Component<DropdownProps, DropdownState> {
         );
     }
 
-    private isInteractedOutside = (e: Event) => {
-        if (!this.isOpened()) {
-            return false;
+    private renderChildren = ({ ref, placement, style, update, isReferenceHidden }: PopperChildrenProps) => {
```
AlekseyManetov, line 183: Maybe gives more specific name, like 'renderDropdownBody'?

uui-components/src/overlays/Dropdown.tsx
```diff
@@ -183,17 +180,39 @@ export class Dropdown extends React.Component<DropdownProps, DropdownState> {
...
+        if (isReferenceHidden && this.state.opened) {
+            setTimeout(() => { this.handleOpenedChange(false); }, 0);
+            return null;
         }
+        return (
+            <div
+                className='uui-popper'
+                ref={ node => {
```
AlekseyManetov, line 191: Can you move it in the separate function too

uui-components/src/overlays/Tooltip.tsx
```diff
@@ -1,23 +1,24 @@
...
     renderContent?(): any;
-    placement?: PopperJS.Placement;
+    placement?: Placement;
     trigger?: 'click' | 'press' | 'hover';
     portalTarget?: HTMLElement;
-    offset?: string;
+    offset?: Options['offset'];
     children?: React.ReactNode;
-    boundaryElement?: Boundary | Element;
+    boundaryElement?: Boundary;
```
AlekseyManetov, line 21: In Popper 2 'Element' can't be passed?

uui-components/src/overlays/Tooltip.tsx
```diff
@@ -129,6 +130,25 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
         </div>;
     }
 
+    private renderChildren = ({ ref, placement, style, arrowProps, isReferenceHidden }: PopperChildrenProps) => {
```
AlekseyManetov, line 133: Maybe gives more specific name, like 'renderTooltipBody'?

uui-components/src/overlays/Tooltip.tsx
```diff
@@ -129,6 +130,25 @@ export class Tooltip extends React.Component<TooltipProps, TooltipState> {
...
+        if (isReferenceHidden && this.state.isOpen) {
```
AlekseyManetov, line 134: Please add comment : 
// Yes, we know that it's hack and we can perform setState in render, but we don't have other way to do it in this case

AlekseyManetov (changes requested):

uui-components/src/overlays/Dropdown.tsx
dsv-sergey: image
Boundary is typed as HTMLElement, which inherits from Element

uui-components/src/overlays/Tooltip.tsx
dsv-sergey: Boundary is typed as HTMLElement, which extends from Element

AlekseyManetov (approved):

PR #10, opened 2021-03-23
Fix scrollbars

dsv-sergey:
[Scrollbars][Loveship][Promo]: updated styles according design
[ScrollManager]: fixed the correctness of work when resizing a window and changing the client width of elements nested in scrollbars


loveship/components/tables/DataTableScrollRow.scss
```diff
@@ -1,4 +1,26 @@
+@use  '../../assets/styles/scss/loveship-colors' as *;
+
 .cell {
     box-sizing: border-box;
     padding: 0 12px;
 }
+
+:global(.uui-table-scroll-row-container) {
+
+    :global(.uui-thumb) {
```
AlekseyManetov, line 10: Can we make it mixin function and use in all cases?

epam-promo/components/tables/DataTableScrollRow.scss
```diff
@@ -1,4 +1,26 @@
...
+
 .cell {
     box-sizing: border-box;
     padding: 0 12px;
 }
+
+:global(.uui-table-scroll-row-container) {
+
+    :global(.uui-thumb) {
+        background: $gray40;
```
AlekseyManetov, line 11: Can we make it mixin function and use in all cases?

AlekseyManetov (changes requested):

PR #16, opened 2021-03-25
Fix/radio input

dsv-sergey:
fixed point display when changing zoom in browsers


epam-promo/components/inputs/RadioInput.scss
```diff
@@ -47,80 +49,143 @@
...
+            :global(.uui-icon) {
+                fill: var(--fill-color-checked-disabled);
             }
         }
+    }
 
+    &:global(.uui-readonly),
+    &:global(.uui-readonly):global(.uui-checked),
+    &:global(.uui-readonly):hover,
+    &:global(.uui-readonly):focus {
```
AlekseyManetov, line 167: Do we have hove and focus styles for .uui-input-label, which we need to change by readonly and disabled mods?

AlekseyManetov (changes requested):

PR #84, opened 2021-07-06
[AnalyticsContext]: implement new AnalyticsListener so that remove AmplitudeClient from uui

okrypets:
[AnalyticsContext]: implement new AnalyticsListener so that remove AmplitudeClient from uui


okrypets:
Not actual

Issue #98, opened 2021-07-22
[New Component]: Scroll Spy


rayetzki:
Scroll Spy component for use cases like Forms, Modal windows, Anchors

PR #101, opened 2021-07-27
[Scroll Spy]: Scroll Spy component implementation

rayetzki:
#98 Scroll Spy component with 3 use cases, that can be found under Sandbox tab


app/src/sandbox/scroll-spy/ScrollSpy.tsx
```diff
@@ -0,0 +1,80 @@
...
+    }, [ref]);
+
+    useEffect(() => {
+        if (observedNodes.length === 0) return;
+        const observer = new IntersectionObserver(entries => {
+            const intersectingElement = entries.find(entry => entry.isIntersecting) as any;
+            setCurrentActive(intersectingElement?.target?.dataset?.spy);
+        }, {
+            ...options,
+            root: options?.root || document.querySelector('body')
```
AlekseyManetov, line 62: I think we will use element which we get from setRef callback to find elements instead of 'root' prop

app/src/sandbox/scroll-spy/ScrollSpy.tsx
```diff
@@ -0,0 +1,80 @@
...
+    initialActive?: string;
+    options?: IntersectionObserverInit;
+    children?: (api: IScrollSpyApi) => ReactNode;
+}
+
+export function useScrollSpy(
+    elements?: IScrollSpyProps['elements'],
+    initialActive?: IScrollSpyProps['initialActive'],
+    options?: IScrollSpyProps['options']
+) : IScrollSpyApi {
```
AlekseyManetov, line 21: You will use IScrollSpyProps interface for params and for ScrollSpyContainer you will extend it with IHasChildren interface

app/src/sandbox/scroll-spy/ScrollSpy.tsx
```diff
@@ -0,0 +1,80 @@
...
+    }, [observedNodes]);
+
+    return {
+        scrollToElement,
+        currentActive,
+        setRef: (selectedRef: HTMLElement) => ref.current = selectedRef,
+    };
+};
+
+export function ScrollSpyContainer({ elements, children } : IScrollSpyProps): ReactNode {
```
AlekseyManetov, line 77: It's don't render any container so we can rename it just to 'ScrollSpy'

app/src/sandbox/scroll-spy/ScrollSpy.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+        observedNodes.forEach(element => element ? observer.observe(element) : null);
+
+        return () => observer.disconnect();
+    }, [observedNodes]);
+
+    return {
+        scrollToElement,
+        currentActive,
+        setRef: (selectedRef: HTMLElement) => ref.current = selectedRef,
```
AlekseyManetov, line 73: Make setRef function via useCallback

AlekseyManetov (changes requested):

app/src/sandbox/scroll-spy/ScrollSpy.tsx
rayetzki: Those sometimes can be different elements. In the Modal and Anchor examples above, there is a section tag that wraps the content, where provided elements ids are queried. But for the Intersection Observer to work properly, the root container must be an element with scroll bars within to track intersection properly. So, I'd propose to leave this option up to user, since it may cause bugs and unexpected behaviours.

PR #119, opened 2021-08-06
[Bugfix]: FIx data cell checkboxes after checkboxes refactor

rayetzki:
FIx data cell checkboxes position

**Before**:
image

**After**:
image


rayetzki:
Will be fixed after checkboxes refactoring

Issue #104, opened 2021-07-29
[Accessibility]: Keyboard Support and correct aria-* mapping


rayetzki:
**Description**
Make sure that UUI Components are fully accessible from keyboard.

Examples:

1. Checkboxes should be focusable by pressing Tab. And by pressing Shift + Tab user can go 1 element behind.
2. Radio Buttons must be selectable by pressing Left-Right arrows.
3. Buttons must be focusable by pressing Tab or Shift Tab.
 
**Describe the solution you'd like**

- [x]  **Alert, Notification card**

    Alert top-level block must have a `role="alertdialog"` or `role="alert"`  and `aria-modal` attribute, only if alert is a modal window. By clicking `Escape`, user should be able to close the alert. Make sure that action buttons within the alert component are accessible from keyboard (Tab Sequence is preserved and buttons are clickable via `Space` or `Enter`
    ****Further reading: [https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/alertdialog.html](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/alertdialog.html)

- [x]  **Avatar Stack**

    Make sure an image container is marked via `role="group"`

- [x]  **Accordion**

    Accordions must be focusable.

    Keyboard support :

    - By pressing `Enter` or `Space` - target accordion must get expanded.
    - `Tab + Shift-Tab` - elements sequence navigation
    - `Down Arrow | Up Arrow` - Up and Down navigation between multiple accordions (if present).

    Assistive technologies use `aria-expanded` attribute to indicate, whether the accordion is expanded or not. If the accordion is disabled, the proper `aria-disabled` should also be set, since there is no native attribute. Screen readers use a combination of `aria-controls ~ id` attribute sequence to correctly map accordion header to accordion body. So, `aria-controls` with a unique id must be set on the header of the accordion, and the same value must be provided for an id attribute on the body wrapper to correctly identify containing elements. As an option, for the screen readers easier to name the accordion blocks, we can also provide a meaningful `aria-labelledby` attribute and pass a `role="region"` to the accordion body. 
    Further reading: [https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html](https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html)

- [x]  **Button (Link)**
[...]

PR #128, opened 2021-08-10
[Separators]: Accessible separators

rayetzki:
As a part of #104 (Separator)

### DIFF
--- a/uui-components/src/layout/flexItems/FlexSpacer.tsx
+++ b/uui-components/src/layout/flexItems/FlexSpacer.tsx
@@ -4,4 +4,4 @@ import { IAdaptiveItem } from '@epam/uui';
 
 export interface FlexSpacerProps extends IAdaptiveItem {}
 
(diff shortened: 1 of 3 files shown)


uui-components/src/layout/flexItems/FlexSpacer.tsx
```diff
@@ -4,4 +4,4 @@ import { IAdaptiveItem } from '@epam/uui';
 
 export interface FlexSpacerProps extends IAdaptiveItem {}
 
-export const FlexSpacer = (props: FlexSpacerProps) => <div { ...props } className={ css.flexSpacer } />;
\ No newline at end of file
+export const FlexSpacer = (props: FlexSpacerProps) => <hr role="separator" { ...props } className={ css.flexSpacer } />;
```
AlekseyManetov, line 7: This is not separator, it flex element which helps to fill space between flex items. Often used to position elements at the beginning and end of a line with space between them.

PR #129, opened 2021-08-10
[FileUpload]: Accessible browse button on File Upload component

rayetzki:
As a part of #104 (File Upload)


AlekseyManetov:
We don't need it here

PR #130, opened 2021-08-10
[Picker Input]: Picker input accessibility

rayetzki:
As a part of #104 (Picker Input)


AlekseyManetov (changes requested):
review for loveship

PR #134, opened 2021-08-10
[Slider Rating]: Slider rating a11y

rayetzki:
As a part of #104 (Slider Rating)

PR #135, opened 2021-08-10
[Star Rating]: Adjust start rating to work from keyboard

rayetzki:
As a part of #104 (Star Rating)

### DIFF
--- a/uui-components/src/inputs/Rating/BaseRating.scss
+++ b/uui-components/src/inputs/Rating/BaseRating.scss
@@ -2,6 +2,10 @@
     position: relative;
     cursor: pointer;
 
+    &:focus, &:focus-visible {
+        outline: none;
+    }
+
     &--readonly {
         cursor: default;
     }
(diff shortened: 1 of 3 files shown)


uui-components/src/inputs/Rating/BaseRating.tsx
```diff
@@ -99,6 +99,14 @@ export class BaseRating extends React.Component<BaseRatingProps<number>, BaseRat
         this.props.onValueChange(this.checkRating(this.getRatingFromWidth(width)));
     }
 
+    onKeyDown(e: React.KeyboardEvent<HTMLDivElement>) {
```
AlekseyManetov, line 102: You don't use this function in component

uui-components/src/inputs/Rating/BaseRating.scss
```diff
@@ -2,6 +2,10 @@
     position: relative;
     cursor: pointer;
 
+    &:focus, &:focus-visible {
```
AlekseyManetov, line 5: there is no outline for this component, because it is not focusable

rayetzki:
Will be merged as a part of GH-134

PR #151, opened 2021-08-17
[Bugfix]: Fix pickerinput focus management

rayetzki:
Previously, focus was manually managed on the picker input wrapper, which lead to tab sequence breaks. Now the tabIndex and focus-blur events are passed directly to the pickertoggler.


rayetzki:
Fixed in #167

rayetzki:
Fixed in #167

PR #167, opened 2021-08-25
[PickerInput]: Fix tab sequence

rayetzki:
Fix tab sequence (Shift-Tab + Tab) once focus is set to PickerInput component.


rayetzki:
Will be merged as a part via ```feature/a11y-pickerinput```

PR #196, opened 2021-09-07
[Fieldset, Radio, Checkbox, Switch]: Pass aria-* attrs for required, …

rayetzki:
Pass aria* attrs for required, disabled states to Radio, Checkbox, Switch
Remove redundant required state from Radio-, CheckboxGroups.


uui-components/src/inputs/RadioInput.tsx
```diff
@@ -52,10 +52,13 @@ export class RadioInput extends React.Component<RadioInputProps, any> {
                 <div className={ cx(uuiElement.radioInput, this.props.value && uuiMod.checked) }>
                     <input
                         type="radio"
-                        checked={ this.props.value }
                         disabled={ this.props.isReadonly || this.props.isDisabled }
+                        aria-disabled={ this.props.isReadonly || this.props.isDisabled }
```
AlekseyManetov, line 56: separate this.props.isReadonly and this.props.isDisabled and aria attrs. Make it in all input components which have readonly and disabled

PR #198, opened 2021-09-07
[PickerInput]: fix size of PickerToggler in size-24 loveship skin

### LINKED ISSUE #188: [Loveship][PickerInput]: Picker change his height after picking/clearing values
# Expected Behavior

The height of picker input should not change after picking/clearing values

# Current Behavior

Picker height changes 

## Steps to Reproduce

1. set picker size to 24
2. set picker selectionMode to 'multi'
image

## Context

UUI3[Loveship] # 4.2.1

okrypets:
fix #188

### DIFF
--- a/loveship/components/pickers/PickerToggler.scss
+++ b/loveship/components/pickers/PickerToggler.scss
@@ -7,23 +7,20 @@
     }
 
     :global(.uui-input) {
-        @include text-size($size - 6px);
+        @include text-size($size);
 
         @if ($size == 24px) {
-            line-height: 18px;
-            font-size: 12px;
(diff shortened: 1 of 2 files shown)


loveship/components/pickers/PickerToggler.scss
```diff
@@ -7,7 +7,7 @@
     }
 
     :global(.uui-input) {
-        @include text-size($size - 6px);
+        @include text-size($size - 6px, 3px);
```
AlekseyManetov, line 10: Make separate specific for toggler vpadding mixin and use it here

PR #214, opened 2021-09-13
[:focus-visible]: Remove :focus-visible

rayetzki:
Remove :focus-visible from components, that rely on it, as it is yet not supported in Safari


pavelspichonak:
@rayetzki could you please revert these changes because Safari already supports `:focus-visible`?

AlekseyManetov:
HI @pavelspichonak, we can't rework this commit because it was created more that 1 year ago, i think we need to create the new one based on our new codebase and also discuss this one more time with our design team.

PR #315, opened 2021-11-23
[DataTable][VirtualList]: Refactor

### LINKED ISSUE #88: Tables layout rework - POC
We need to try new approaches to tables, to improve performance:
- css grid
- position: sticky for fixed columns (and, probably, header)
- new https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API can speed-up VirtualList. Example: https://github.com/researchgate/react-intersection-list

rayetzki:
1. Extract virtualization functionality from VirtualList primarily to useVirtual hook to re-use in both VirtualList and DataTable.
2. Refactor components to use Scrollbars from @epam/uui-components, rather than from package directly.
3. Refactor DataTable to base class.
4. Remove ScrollManager from table-related components together with DataTableScrollRow that used to rely on it.
5. Refactor layout to use native scrollbars.
6. Refactor table shadows to rely on IntersectionObserver rather than scroll position.
7. Improve typings on table-related components.

Closes #88


epam-promo/components/tables/types.ts
```diff
@@ -16,6 +16,7 @@ export interface DataTableMods {
     size?: ControlSize;
     border?: 'none' | 'gray30';
     headerTextCase?: 'upper' | 'normal';
+    shadow?: 'dark' | 'white';
```
AlekseyManetov, line 19: remove, make default by design

uui-components/src/layout/ScrollBars.scss
```diff
@@ -21,11 +21,48 @@
         background: linear-gradient(to top, rgb(0, 0, 0) 0%, rgba(0, 0, 0, 0) 100%);
     }
 
+    &:global(.uui-scroll-shadow-top) {
```
AlekseyManetov, line 24: check do we need shadows here

uui-components/src/table/DataTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+export function DataTable<TItem, TId>({
+    rowsCount,
+    value,
+    onValueChange,
+    onScroll,
+    columns,
+    ...props
+}: DataTableProps<TItem, TId>) {
+    const { listRef, scrollbarsRef, offsetY, handleScroll, estimatedHeight } = useVirtualList<HTMLDivElement>({
```
AlekseyManetov, line 37: move useVirtualLust hook usage and shadows logic to the VirtualList component. Remove DataTable from uui-components, just leave in skins

PR #347, opened 2021-12-14
[LazyDataSource]: selectAll attribute

### LINKED ISSUE #146: [LazyDataSource]: add possibility to disable selectAll behavior
**Description**
In some cases you don't need SelectAll behavior on lazy lists, for example if you have list with too much items for upload. 

**Describe the solution you'd like**
We need to add some configuration to disable SelectAll, i think it will coma as LazyDataSource option, for example - "disalbeSelectAll"

rayetzki:
Closes #146

### DIFF
--- a/uui/data/processing/views/ArrayListView.ts
+++ b/uui/data/processing/views/ArrayListView.ts
@@ -171,7 +171,9 @@ export class ArrayListView<TItem, TId, TFilter = any> extends BaseListView<TItem
             row.isFocused = this.value.focusedIndex === index;
         });
 
-        if (all.checkableCount > 0) {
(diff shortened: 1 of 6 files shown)


uui/data/processing/views/ArrayListView.ts
```diff
@@ -171,7 +171,9 @@ export class ArrayListView<TItem, TId, TFilter = any> extends BaseListView<TItem
             row.isFocused = this.value.focusedIndex === index;
         });
 
-        if (all.checkableCount > 0) {
+        const isSelectAllSwitched = this.props.selectAll == undefined ? true : this.props.selectAll;
```
AlekseyManetov, line 174: don't copy-paste this for all views, move this to the BaseListView. And may be it better to name this - 'isSelectAllEnabled'?

PR #417, opened 2022-02-02
Feature/uui-v

dsv-sergey:


epam-promo/package.json
```diff
@@ -13,6 +13,7 @@
   "dependencies": {
     "@epam/assets": "4.5.3",
     "@epam/uui": "4.5.3",
+    "@epam/uui-v": "4.5.3",
```
AlekseyManetov, line 16: remove dependency on @epam/uui-v

uui-v/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@epam/uui-v",
+  "version": "4.5.3",
+  "description": "EPAM UUI-V components set",
+  "author": "EPAM",
+  "license": "MIT",
+  "main": "index.js",
```
AlekseyManetov, line 7: add   `"private": true,`, because we don't plan to publish this package to npm now

AlekseyManetov (changes requested):

PR #420, opened 2022-02-04
[Assets Doc]: Fix misleading warning

rayetzki:
For locked and unlocked states, if the icon exists in any possible sizes, do not show misleading warning.


app/src/docs/assets/Icons.doc.tsx
```diff
@@ -125,24 +130,26 @@ export class IconsDoc extends React.Component {
...
+                <FlexRow size='24' vPadding='24'>
+                    <TextInput value='Some text' size={ this.state.controlSize } onValueChange={ () => {} } icon={ icon } />
                 </FlexRow>
             </FlexCell>
         );
     }
 
     checkValidSize = () => {
-        if (typeof recommendedSizes[this.state.selectedIcon.size] === 'undefined') {
+        if (this.state.selectedIcon.name) {
```
AlekseyManetov, line 150: why do we have here condition with `selectedIcon.name`? We need check about if we have icon with only invalid sizes, we don't need to show this warning.

Issue #462, opened 2022-02-24
[LabeledInput]: info tooltip doesn't work


JuliaMV:
# Expected Behavior

Info tooltip should appear on hover

# Current Behavior

Tooltip doesn't appear

## Steps to Reproduce
https://uui.epam.com/documents?category=components&id=labeledInput&mode=propsEditor&skin=UUI4_promo
![2022-02-24_13h56_59](https://user-images.githubusercontent.com/37875890/155514883-26293825-f3ba-4aa9-8da7-dc73d43e40f4.png)

## Context

* UUI version:  4.6.0-alpha.1
* Browser name: any
* Operating System: any

PR #464, opened 2022-03-02
[LabeledInfo]: show tooltip for info

JuliaMV:
to fix https://github.com/epam/UUI/issues/462. I don't know the reason of issue but it works with this fix.


AlekseyManetov:
fixed by commit - https://github.com/epam/UUI/commit/6346ce88b60debb635519f0f0caea96d137b6e67

PR #469, opened 2022-03-11
Added "align", "disableArrows" and "mode" props to NumericInput

Qwerias:


uui-components/src/inputs/NumericInput.tsx
```diff
@@ -131,33 +134,36 @@ export class NumericInput extends React.Component<NumericInputProps, NumericInpu
             >
                 <input
                     type="number"
-                    className={ cx(uuiElement.input, this.props.inputCx) }
+                    className={ cx(uuiElement.input, this.props.inputCx, this.props.disableArrows, this.props.align === "right" && css.alignRight) }
```
AlekseyManetov, line 137: do we need here `this.props.disableArrows`?

AlekseyManetov (changes requested):

PR #521, opened 2022-04-17
[useForm]: Add type to handleSave's return value in useForm

alex9pro:
`handleSave` function returns Promise, but in `useForm`'s `save` method we have `void` as return type.

We have plans use save to catch Promise.reject after validation, for now we need to cast type:

```tsx
((form.save() as unknown) as Promise<void>).catch(() => {
    scrollManager.scrollToInvalid();
});
```

With this fix we will get rid of annoying type casting ☺

```tsx
form.save().catch(() => {
    scrollManager.scrollToInvalid();
});
```

### DIFF
--- a/uui-core/data/forms/Form.tsx
+++ b/uui-core/data/forms/Form.tsx
@@ -20,7 +20,7 @@ export interface FormProps<T> {
 }
 
 export interface RenderFormProps<T> extends IEditable<T>, ICanBeInvalid {
-    save(): void;
+    save(): Promise<void>;
     undo(): void;
     redo(): void;
     revert(): void;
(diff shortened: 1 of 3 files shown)


AlekseyManetov:
Hi @alex9pro! Could you please describe for which case use need promise here? If you need to handle server response, for example to show notification, you can use onSucces and onErrror callback in useForm hook.

alex9pro:
Hi @AlekseyManetov! I'd like call `modalProps.success` method in Modal component:

1. Close Modal on success
```tsx
const Modal = ({ children, success, form }) => {
  // I want have common modal component for form and close modal via success prop in this component
  const handleSave = () => form.save().then(() => success(true));

  return (
    <ModalWindow>
        {children}
        <Button caption="Save" onClick={handleSave} />
    </ModalWindow>
  );
}
```

```tsx
const ModalUser = (modalProps) => {
  const form = useForm();
  
  return <Modal {modalProps} form={form}></Modal>
}
```

2. Scroll to fields with errors

I need to have possibility scroll to fields with errors after save.

As I see I can use save().catch because if form invalid Promise.reject returns in handleSave

image

Issue #522, opened 2022-04-18
Getting error from fresh installation - TS2769 & TS2786


Aravin:
# Prerequisites

- [Yes ] I am running the latest version
- [ Yes] I checked the documentation and found no answer
- [ Yes] I checked to make sure that this issue has not already been filed

# Description

I have installed the fresh version of UUI and started the service. I could see 2 errors from the App.tsx and Index.tsx file

## Steps to Reproduce
1. run yarn create react-app my-app --template @epam/uui
2. cd my-app
3. yarn start 

What is the current behavior?

Getting below errors

```
Compiled with problems:X

ERROR in src/App.tsx:12:39

TS2769: No overload matches this call.
  Overload 1 of 2, '(props: (RouteProps<"/", {}> & OmitNative<{}, keyof RouteProps<string, { [x: string]: string | undefined; }>>) | Readonly<RouteProps<"/", {}> & OmitNative<...>>): Route<...>', gave the following error.
    Type 'typeof MainPage' is not assignable to type 'ComponentType<RouteComponentProps<any, StaticContext, unknown>> | ComponentType<any> | undefined'.
      Type 'typeof MainPage' is not assignable to type 'ComponentClass<RouteComponentProps<any, StaticContext, unknown>, any>'.
        Types of property 'contextType' are incompatible.
          Type 'React.Context<any> | undefined' is not assignable to type 'import("/Users/aravin/.Trash/interview-feedback/interview-feedback/node_modules/@types/react-router/node_modules/@types/react/index").Context<any> | undefined'.
            Type 'React.Context<any>' is not assignable to type 'import("/Users/aravin/.Trash/interview-feedback/interview-feedback/node_modules/@types/react-router/node_modules/@types/react/index").Context<any>'.
              The types of 'Provider.propTypes' are incompatible between these types.
                Type 'React.WeakValidationMap<React.ProviderProps<any>> | undefined' is not assignable to type 'import("/Users/aravin/.Trash/interview-feedback/interview-feedback/node_modules/@types/react-router/node_modules/@types/react/index").WeakValidationMap<import("/Users/aravin/.Trash/interview-feedback/interview-feedback/node_modules/@types/react-router/node_modules/@types/react/index").ProviderProps<any>> | undefined'.
[...]

PR #532, opened 2022-04-28
Fix/i522 update uui cra template

dsv-sergey:
Fixed issue #522


app/src/docs/examples/contexts/AnalyticsContextBase.example.tsx
```diff
@@ -29,7 +29,7 @@ class AmplitudeListener implements IAnalyticsListener {
 const AnalyticsContextBase: React.FC = () => {
     const loadAppContext = useCallback(() => Promise.resolve(), []);
 
-    const onInitCompleted = useCallback((context) => {
+    const onInitCompleted = useCallback((context: any) => {
```
Aravin, line 32: can we use a specific type?

Aravin (approved):

Issue #537, opened 2022-05-04
[Multiswitch]: Add 2nd (light) context


Yaroslav-Zinkov:
**Description**
Please add a new version of the component

**Describe the solution you'd like**
UUI-3 guidline
https://www.figma.com/file/M5Njgc6SQJ3TPUccp5XHQx/UUI3-(Components)?node-id=6022%3A185850
UUI-4 guideline
https://www.figma.com/file/UyChXPLmyv5zMrOU37KdUL/UUI4-(Components)?node-id=13057%3A230008

<img width="687" alt="image" src="https://user-images.githubusercontent.com/89978100/166646917-8ee929ed-8a6c-46bd-a268-20aefe1d499a.png">

vik753:
In loveship theme for this design using props color='night600'. It temporary solution before we will made new button version with right design.

PR #550, opened 2022-05-19
Feature/multiswitch light context #537

vik753:


vik753:
In loveship theme for this design using props color='night600'. It temporary solution before we will made new button version with right design.

PR #554, opened 2022-05-24
fixed area-expended value in the DataTableRow

vik753:
fixed area-expended value in the DataTableRow

### DIFF
--- a/uui-components/src/table/DataTableRow.tsx
+++ b/uui-components/src/table/DataTableRow.tsx
@@ -36,8 +36,8 @@ export class DataTableRow<TItem, TId> extends Component<DataTableRowProps<TItem,
                 rawProps={ {
                     ...params.eventHandlers,
                     role: 'row',
-                    'aria-expanded': this.props.isFolded == undefined ? undefined : this.props.isFolded,
-                    ...(this.props.isSelectable && { 'aria-selected': this.props.isSelected } )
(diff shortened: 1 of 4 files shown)


uui-components/src/table/DataTableRow.tsx
```diff
@@ -36,8 +36,8 @@ export class DataTableRow<TItem, TId> extends Component<DataTableRowProps<TItem,
                 rawProps={ {
                     ...params.eventHandlers,
                     role: 'row',
-                    'aria-expanded': this.props.isFolded == undefined ? undefined : this.props.isFolded,
-                    ...(this.props.isSelectable && { 'aria-selected': this.props.isSelected } )
+                    'aria-expanded': !this.props.isFolded ?? undefined,
```
vasilii-kovalev, line 39: It seems this code is not fully correct:
* `isFolded === undefined` → `aria-expanded: true`
* `isFolded === false` → `aria-expanded: true`
* `isFolded === true` → `aria-expanded: undefined`

I suppose the initial solution was correct, except there should be `!` symbol added before `this.props.isFolded`.

CC @AlekseyManetov

vasilii-kovalev:
By the way, to link a merge request to issues they close, [some special words](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) can be used. That would make the work on the issues more transparent.

PR #556, opened 2022-05-25
Fix the link to license file in `README.md`

vasilii-kovalev:
The link to the license file in the root `README.md` had a typo. Now it is fixed.
The case type of the file’s extension was also changed so that the link meet the file’s full name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,4 +89,4 @@ This would open the uui.epam.com web-site locally, in watch mode.
 
 ## License
 
-[MIT](./LICENCE.MD)
+[MIT](./LICENSE.md)


vasilii-kovalev:
According to the [contributing guide](https://github.com/epam/UUI/blob/main/CONTRIBUTING.md#creating-a-pull-request) (point 7), it is necessary to add a description of the change to `changelog.md`. But:
1. I don’t see this being done in other pull requests
2. Not sure if it is worth to mention

Should I do that?

AlekseyManetov:
Thanks for contribution. I guess for such change which is not affect our packages, you don't need to update the changelog.md

PR #557, opened 2022-05-26
fixed NoResultBlock into center of the table layout (promo and loveship)

vik753:


epam-promo/components/tables/DataTable.scss
```diff
@@ -37,4 +37,16 @@
         background: linear-gradient(to bottom, rgba(0, 0, 0, 0.05) 0%, rgba(0, 0, 0, 0) 100%);
         transition: display 200ms;
     }
+}
+
+.no-results {
+    position: sticky;
+    left: 0;
+    height: calc(100% - 36px);
```
vasilii-kovalev, line 45: Since the `div` with this class wraps the custom `props.renderNoResultsBlock` component, `36px` won't be possible to override easily. If I understand correctly, this value is equal to the header cells' height, which can be adjusted on the user's side. Would be great to provide an ability to override this value via custom variables, a non-truncated class name, etc.

vasilii-kovalev: The best option here would be to remove the wrapper `div` and pass this class as a property to `props.renderNoResultsBlock` (and the default `DataTableNoResults` component) instead, allowing users to combine it with their local classes.

Issue #562, opened 2022-05-30
[DropSpot] - isDragStart works incorrectly in case when user drag&drop file out of drag area 


anna-kiranova:
# Prerequisites

Please answer the following questions for yourself before submitting an issue. **YOU MAY DELETE THE PREREQUISITES SECTION.**

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

# Description
isDragStart does not work correctly when a user starts dragging a file but mistakenly releases it outside the area (The file is opened in a new tab and DropSpot stays in the "dragging" mode. After that, a file can be attached but isDragStart for next files has a value "false", so a user can't see the "dragging" mode at all) 

## Steps to Reproduce

1. Choose a file and start dragging it into the drag area.
2. Release a file outside of the drag area.
3. Close the opened new tab with your file.
4. See that state of the drag area is in "dragging" mode.
5. Drag a new file into the area.
6. Try to drag another file and see that "dragging" mode doesn't work anymore. 

## Actual result

![2022-05-30_12h18_35](https://user-images.githubusercontent.com/35148952/170982186-f82019f7-53e7-4948-84ba-49a5c2858b61.gif)
![2022-05-30_12h29_39](https://user-images.githubusercontent.com/35148952/170982207-936521cc-0d1c-48bd-89c0-5a1e86521603.gif)

## Expected result

I expect that isDragStart changes to "false" in case of a user's mistake (when the file is released outside of the area) and works properly when a user tries to attach the next files after the mistake.

# Environment

The issue is reproducible on UUI.

PR #570, opened 2022-06-01
Bugfix/drop spot incorrect drag&drop #562

vik753:


anna-kiranova:
@vik753 This solution doesn't fix the problem I described. I want to highlight the drop area on "isDragStart" event and you didn't fix it :)
Moreover, the "isDragOver" has a big problem with "blinking". See the video:

https://user-images.githubusercontent.com/35148952/171646894-5ae7f7e4-154c-40e5-8de9-9618a62a955e.mp4

PR #739, opened 2022-10-02
Improve issue templates

vasilii-kovalev:
* Turn all the headers into level 2 ones (because h1 is already reserved by the issue's title, and a page should have only one)
* Unify words case in the bug report template:
  * Steps to Reproduce → Steps to **r**eproduce
  * Operating System → Operating **s**ystem
* Rename `bug_report.md` to `bug-report.md`, so that the naming be consistent with other template names (for example, `feature-request`). As far as I understand, it will require changes in the repository's settings to support the new template's name


vasilii-kovalev:
Done.

PR #752, opened 2022-10-09
fixed @ CVE-2021-24033

mik-patient:
## Descriptions Overview
react-dev-utils prior to v11.0.4 exposes a function, `getProcessForPort`, where an input argument is concatenated into a command string to be executed. This function is typically used from react-scripts (in Create React App projects), where the usage is safe. Only when this function is manually invoked with user-provided values (ie: by custom code) is there the potential for command injection. If you're consuming it from react-scripts then this issue does not affect you.

CVE-2021-24033
`CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:L/I:L/A:L`
GHSA-5q6m-3h65-w53x

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -62,19 +62,19 @@
     "@arrows/error" "^1.0.2"
     fast-deep-equal "^3.1.3"
 
-"@babel/code-frame@7.8.3":
-  version "7.8.3"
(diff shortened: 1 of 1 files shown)


AlekseyManetov:
Hi, we use it from react-scripts, so as you say above this will not affect us

PR #753, opened 2022-10-09
fix @ CVE-2021-3801

mik-patient:
## Descriptions
Prism is a syntax highlighting library. The prismjs package is vulnerable to ReDoS (regular expression denial of service). An attacker that is able to provide a crafted HTML comment as input may cause an application to consume an excessive amount of CPU.

CVE-2021-3801
`CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:N/I:N/A:H`
GHSA-hqhp-5p83-hx96


AlekseyManetov:
We will update `prismjs` package into our package.json. Changing it into yarn.lock will be overwritten after the new packages installation.

PR #754, opened 2022-10-09
fix @ CVE-2022-29078

mik-patient:
## Description 
The ejs (aka Embedded JavaScript templates) package 3.1.6 for Node.js allows server-side template injection in settings[view options][outputFunctionName]. This is parsed as an internal option, and overwrites the outputFunctionName option with an arbitrary OS command (which is executed upon template compilation).

CVE-2022-29078
`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H`
GHSA-phwq-j96m-2c2q


AlekseyManetov:
We will update `ejs` package into our package.json. Changing it into `yarn.lock` will be overwritten after the new packages installation. Thanks.

PR #858, opened 2022-12-06
Feature/add delay for dropdown open close

RomSribn:
- add open and close delay to the dropdown;
- display delayed dropdowns at the documents page;
- auto generate code files for delayed dropdowns.

### DIFF
--- a/app/src/docs/examples/dropdown/CloseOpenDelayModifiers.example.tsx
+++ b/app/src/docs/examples/dropdown/CloseOpenDelayModifiers.example.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Dropdown, DropdownBodyProps } from '@epam/uui-components';
 import { Button, Text, FlexRow, DropdownContainer, FlexCell } from '@epam/promo';
-import * as css from './HandleStateExample.scss';
+import css from './HandleStateExample.scss';
 import { Avatar, IDropdownToggler } from '@epam/uui';
 
 export default function CloseOpenDelayModifiersExample() {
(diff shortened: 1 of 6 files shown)


app/src/docs/examples/dropdown/CloseOpenDelayModifiers.example.tsx
```diff
@@ -0,0 +1,169 @@
+import React from 'react';
+import { Dropdown, DropdownBodyProps } from '@epam/uui-components';
+import { Button, Text, FlexRow, DropdownContainer, FlexCell } from '@epam/promo';
+import * as css from './HandleStateExample.scss';
```
siarheiyelin, line 4: could you please use default export instead? named exports will be disabled soon.
`import css from './HandleStateExample.scss'`

app/src/docs/examples/dropdown/CloseOpenDelayModifiers.example.tsx
```diff
@@ -0,0 +1,169 @@
...
+                <Dropdown
+                    renderBody={ (props) => renderDropdownBody(props) }
+                    renderTarget={ (props: IDropdownToggler) => <Button caption="Hover to open (open delay)" { ...props } /> }
+                    openOnHover={ true }
+                    openDelay={ 500 }
+                    closeOnMouseLeave='toggler'
+                />
+            </FlexRow>
+            <FlexRow vPadding='12' spacing='12' >
+                <Dropdown
```
siarheiyelin, line 76: it seems like all 3 elements are identical. Not a big deal, but I would propose to render it like this to avoid code dublication:
```

{
Array(3).fill(null).map((e, i) => (
            <Dropdown key={ i }  renderBody={ (props) => renderDropdownBody(props) }
                renderTarget={ (props: IDropdownToggler) => <Button caption="Hover to open (open delay)" { ...props } /> }
                openOnHover={ true } openDelay={ 500 } closeOnMouseLeave='toggler'
            />
        ))
}

```

uui-components/src/overlays/Dropdown.tsx
```diff
@@ -127,11 +131,16 @@ export class Dropdown extends React.Component<DropdownProps, DropdownState> {
...
     private handleMouseEnter = (e: Event) => {
-        this.handleOpenedChange(true);
+        this.setOpenDropdownTimer();
     }
 
     private handleMouseLeave = (e: MouseEvent) => {
-        this.handleOpenedChange(false);
+        if (this.props.openDelay && !this.props.closeDelay) {
+            this.handleOpenedChange(false);
+            this.clearOpenDropdownTimer();
```
siarheiyelin, line 140: It seems "clearOpenDropdownTimer" should be always executed without conditions, because "open" timer is always set on mouse hover (even if the delay = 0, it's still asynchronous and not executed immediately). I.e. I would rewrite it like this:

```
private handleMouseLeave = (e: MouseEvent) => {
    this.clearOpenDropdownTimer();
    if (!this.props.closeDelay) {
        this.handleOpenedChange(false);
    } else {
        this.setCloseDropdownTimer();
    }
}
```

uui-components/src/overlays/Dropdown.tsx
```diff
@@ -63,6 +66,7 @@ export class Dropdown extends React.Component<DropdownProps, DropdownState> {
         opened: this.props.value || false,
         bodyBoundingRect: { y: null, x: null, height: null, width: null },
         closeDropdownTimerId: null,
+        openDropdownTimerId: null,
```
siarheiyelin, line 69: I think it's a good idea not to set openDropdownTimerId to components state and move it to instance variable instaed. So that: 
- we avoid unnecessary re-renders when timer id is changed
- we avoid potential issues with because setState is asynchronous and the changes are not applied immediately.

```
private openDropdownTimerId: NodeJS.Timeout = null;

```

app/src/docs/examples/dropdown/CloseOpenDelayModifiers.example.tsx
```diff
@@ -0,0 +1,57 @@
...
+                    openOnHover={ true }
+                    openDelay={ 500 }
+                    closeOnMouseLeave='toggler'
+                />
+            </FlexRow>
+    );
+
+    return (
+        <div>
+            { Array(6).fill(null).map((e, i) => <DropdownTextRow />) }
```
siarheiyelin, line 54: key is missing: `<DropdownTextRow key={i} />)`

siarheiyelin (approved):

PR #873, opened 2022-12-10
Add `rel` property to `ICanRedirect` interface

### LINKED ISSUE #857: [DropdownMenuButton] There is no way to add `rel` attribute
## Now
We have a security issue:
Using target without correctly setting the "rel" attribute, or disassociating the new window from its parent, is an unsafe way of opening a new window. 

## To Do
How to avoid it:

  * Do not set the "target" attribute (with any value) for links created by users unless required.
  * If required, when using the "target" attribute, also set the "rel" attribute as “noopener noreferrer”:
      + “noopener” for Chrome and Opera
      + “noreferrer” for Firefox and old browsers

## Describe the solution you'd like (optional)

I suggest to add rel property to `ICanRedirect` interface

svikhristyuk:
Closes #857

### DIFF
--- a/uui-core/src/types/props.ts
+++ b/uui-core/src/types/props.ts
@@ -45,6 +45,9 @@ export interface ICanRedirect {
 
     /** Controls where the link should be opened */
     target?: '_blank';
+
+    /** Defines the relationship between a linked resource and the current document */
+    rel?: string;
 }
 
 /** Component can be disabled */


AlekseyManetov:
It's not enough to add `rel` property only for interface, we also need to provide it on link nodes inside components. I will close this PR and we will add `rel` attribute by default for all links with target: '_blank' in feature release

PR #885, opened 2022-12-20
[Draft]Feature/tooltips examples rework

dzevakov:
- tooltips examples rework
- added delay property, to display the tooltip with a delay
- maxWidth property, to set max width of the tooltip


epam-promo/components/overlays/Tooltip.scss
```diff
@@ -7,7 +7,6 @@
         font-size: 14px;
         font-family: $font-sans;
         box-shadow: 0 1px 4px 0 rgba(44, 47, 60, 0.05), 0 2px 18px 0 rgba(44, 47, 60, 0.03);
-        max-width: 300px;
```
siarheiyelin, line 10: It seems, the same change needs to be applied to "loveship" skin as well: loveship/components/overlays/Tooltip.scss

siarheiyelin (approved):

epam-promo/components/overlays/Tooltip.scss
dzevakov: Thank you for your attention) Fixed

PR #899, opened 2022-12-23
[Avatar]: handle error on loading image and shows stub

Svetzayats:
For what: improve Avatar component, add handling of errors. 

Current behaviour: if there is incorrect path in Avatar (not undefined) -> Avatar shows broken image, not stub. 
After improvement: if there is incorrect path in Avatar -> Avatar shows stub. 

In case when developer pass rawProps with his own handler for onError -> we don't show stub, handling of error on developer side.

### DIFF
--- a/epam-promo/components/widgets/__tests__/Avatar.test.tsx
+++ b/epam-promo/components/widgets/__tests__/Avatar.test.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Avatar } from '../Avatar';
 import renderer from 'react-test-renderer';
+import {render, screen, fireEvent} from '@testing-library/react'
 
 describe('Avatar', () => {
     it('should be rendered correctly', () => {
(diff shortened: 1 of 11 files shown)


epam-promo/components/widgets/__tests__/Avatar.test.tsx
```diff
@@ -25,6 +26,40 @@ describe('Avatar', () => {
             />)
             .toJSON();
         expect(tree).toMatchSnapshot();
+
+    });
```
AlekseyManetov, line 30: Please move this tests to uui-components package, because this logic implements there. In skins we normally have only snapshots, but component logic we tests in that package we it implemented.

uui-components/src/widgets/Avatar.tsx
```diff
@@ -20,15 +20,23 @@ export interface AvatarProps extends IHasCX, IHasRawProps<React.ImgHTMLAttribute
 }
 
 const AvatarComponent = (props: AvatarProps, ref: React.ForwardedRef<HTMLImageElement>) => {
+    const [errorMode, setErrorMode] = React.useState<boolean>(false);
```
AlekseyManetov, line 23: rename to [isError, setIsError]

AlekseyManetov (changes requested):

Svetzayats:
Fixed comments

PR #928, opened 2023-01-06
[ColumnsConfigurationModal]: fixed column with `fix` checkbox and pin behavior.

Kunzetsov:
## Summary

- Removed disabling of a checkbox if a column has a `fix` property.
- Fixed the problem with pinning the column after unpinning, if it has `fix` property in the column config.

#### Before: 

https://user-images.githubusercontent.com/22456368/211057844-530665d2-6be8-43fb-8d00-6cc8eb8726b7.mov

#### After: 

https://user-images.githubusercontent.com/22456368/211058491-5faa9246-b0ad-4fd6-8e0f-909045565d8e.mov

#### Before:

https://user-images.githubusercontent.com/22456368/211058043-a19c315a-6dfa-4ae5-b494-b495d2933fea.mov

#### After:

https://user-images.githubusercontent.com/22456368/211057197-c4347c8f-4025-4e68-9615-73ef15e3c59a.mov


uui-core/src/helpers/applyColumnsConfig.ts
```diff
@@ -28,11 +28,12 @@ export const getColumnsConfig = <TItem, TId>(columns: DataColumnProps<TItem, TId
...
     columns.forEach(column => {
         const order = getOrderBetween(prevOrder, null);
-        const columnConfig = config[column.key] || {};
+        const hasPrevColumnConfig = !!config[column.key];
+        const columnConfig = hasPrevColumnConfig ? config[column.key] : {};
 
         config[column.key] = {
             width: columnConfig.width || column.width,
-            fix: columnConfig.fix || column.fix,
+            fix: hasPrevColumnConfig ? columnConfig.fix : column.fix,
```
AlekseyManetov, line 36: May be we can simplify it to `config[column.key]?.fix || column.fix`?

AlekseyManetov (changes requested):

uui-core/src/helpers/applyColumnsConfig.ts
Kunzetsov: We can't. If `config[column.key].fix` will be undefined, it will be rewritten by `column.fix`, which is the source of the existing bug.

NatalliaAlieva:
Released in 4.10.0 version.

Issue #916, opened 2023-01-02
[NumericInput] formatOptions prop works incorrect


mari-shakhanava:
## Steps to Reproduce

Please provide detailed steps for reproducing the issue.

1. There is NumericInput with prop formatOptions={{ minimumFractionDigits: 2, maximumFractionDigits: 2 }}
2. Try to enter 100.56 and then click '6' one more time

## Actual result

Total value increases on 0.01 each time by pressing '6' button (or any number >5).

## Expected result

Total value doesn't change

## Environment
* UUI version: latest
* Browser name: Chrome
* Operating System: Windows

AlekseyManetov:
fixed in 4.10.0 version

Kuznietsov:
Unfortunately, the fix, which was provided, is not providing a good user experience, so it has been reverted. During the investigation of the possible solutions, we haven't found one, which will provide high-precision floating point numerical values support, except the one with string values without conversion from string to a number. Unfortunately, it will not match the expectations of the library's users, because before they were facing problems with working with strings.

cc @AlekseyManetov

PR #933, opened 2023-01-10
[NumericInput]: fixed formatOptions prop works incorrect 

Kunzetsov:
### Summary
Closes: https://github.com/epam/UUI/issues/916

Replaced `toFixed` with a function, which is not rounding up the number.

#### Before:

https://user-images.githubusercontent.com/22456368/211508939-d26703f8-ce58-4318-8459-4b7e48cd3ec9.mov

#### After:

https://user-images.githubusercontent.com/22456368/211508773-935e5848-5ad2-4eef-bec4-dbf9e0b1d8dc.mov


uui-components/src/inputs/NumericInput.tsx
```diff
@@ -60,6 +60,12 @@ const getFractionDigits = (formatOptions: Intl.NumberFormatOptions) => {
     return maximumFractionDigits;
 };
 
+const toFixedWithoutRoundingUp = (value: number, fractionDigits: number) => {
```
AlekseyManetov, line 63: Let's move this function into `uui-core/src/helpers/numericInputCalculations.ts `and also add tests for it

AlekseyManetov (changes requested):

PR #1009, opened 2023-01-31
chor: added --openssl-legacy-provider flag to yarn start

tarekselem:


AlekseyManetov:
Why we need this change?

tarekselem:
@AlekseyManetov  Does it work fine with out that flag?

siarheiyelin:
@tarekselem This issue is no longer actual, since we have recently migrated to Webpack 5. However the changes are currently committed only to "develop" branch. I.e. it's not merged to "main" yet.  In the meantime, please use "develop" branch instead.

PR #1012, opened 2023-02-01
[PickerInput][LazyListView]: enabled search over children

Kuznietsov:
### Summary

Enabled search over childer in `PickerInput` if a data source is `LazyListView`.

https://user-images.githubusercontent.com/22456368/216074028-5fe0bd03-1a30-409b-862a-74afdbb56764.mov


uui-core/src/data/processing/views/Tree.ts
```diff
@@ -533,6 +533,17 @@ export class Tree<TItem, TId> {
         }
     }
 
+    public getParents(id: TId) {
+        const parentIds = this.getParentIdsRecursive(id);
+        const parents = parentIds.reduce<TItem[]>((acc, parentId) => {
+            if (!this.byId.has(parentId)) {
+                return acc;
+            }
+            return [...acc, this.byId.get(parentId)];
```
jakobz, line 542: The [...acc, something] clones the array. While this may be not significant in this case - there's not many parents, in general such pattern for .reduce() is considered as bad.

uui-core/src/data/processing/views/LazyListView.ts
```diff
@@ -316,8 +326,13 @@ export class LazyListView<TItem, TId, TFilter = any> extends BaseListView<TItem,
             for (let n = 0; n < ids.length; n++) {
                 const id = ids[n];
                 const item = this.tree.getById(id);
-                const row = this.getRowProps(item, index, parents);
+                const hasNotIncludedParents = this.props.getParentId?.(item) && !parents.length;
```
jakobz, line 329: I'd rather always compute parents via the Tree. This shouldn't be slower, and we'll simplify the code by removing additional code paths.

uui-core/src/data/processing/views/LazyListView.ts
```diff
@@ -278,13 +280,21 @@ export class LazyListView<TItem, TId, TFilter = any> extends BaseListView<TItem,
...
 
-        } catch(e) {
+        } catch (e) {
             // TBD - correct error handling
             console.log("LazyListView: Error while loading items");
             return { isUpdated: false, isOutdated: false, tree: loadingTree };
         }
     }
-
+    private collectNotIncludedParents(id: TId, item: TItem): DataRowProps<TItem, TId>[] {
```
jakobz, line 289: I believe we need to re-factor the getRowProps. There's a problem:
- we create a tons for DataTableRows
- to put them into getRowProps(_ , _ , parents)
- so the getRowProps gets several of props from the DataRowProps to put into the DataRowProps.path

uui-core/src/data/processing/views/Tree.ts
Kunzetsov: Ok, that makes sense. I just knew that it is a reduced amount of parents here. Will change.

AlekseyManetov (changes requested):

PR #1087, opened 2023-03-02
[Tree] patch

Kuznietsov:
### Summary
Added patch function to a tree for the purpose of updating the existing tree with incoming items.


uui-core/src/data/processing/views/Tree.ts
```diff
@@ -223,47 +215,71 @@ export class Tree<TItem, TId> {
...
         const newById = this.cloneMap(this.byId);
         const newByParentId = this.cloneMap(this.byParentId); // shallow clone, still need to copy arrays inside!
 
-        itemsToAdd.forEach((item) => {
+        let isPatched = false;
+        items.forEach((item) => {
             const id = this.getId(item);
             const existingItem = this.byId.get(id);
+            const parentId = this.getParentId(item);
+            let list = [...(newByParentId.get(parentId) ?? [])];
```
jakobz, line 243: Probably we can avoid cloning the list for every item, even if it's not inserted/deleted?

uui-core/src/data/processing/views/treeHelpers.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const pasteNewItemIntoList = <TItem, TId>(
+    id: TId,
+    newItem: TItem,
+    byId: IMap<TId, TItem>,
+    list: TId[],
+    comparator: (newItem: TItem, existingItem: TItem) => number,
+) => {
+    const listWithNewItem: TId[] = [];
+    list.forEach((itemId) => {
```
jakobz, line 35: Can we simplify this, and avoid o(n^2)? Probably we can
1. find the first index, where newItem is greater than existing (by comparator)
2. insert item into this position

uui-core/src/data/processing/views/treeHelpers.ts
```diff
@@ -0,0 +1,73 @@
...
+        list.splice(foundIndex, 1);
+    }
+
+    return list;
+};
+
+const pasteNewItemIntoList = <TItem, TId>(
+    id: TId,
+    newItem: TItem,
+    byId: IMap<TId, TItem>,
```
jakobz, line 30: Maybe just pass the Tree as arguments here and in other helpers, instead of passing maps? This would simplify parameters, and don't break encapsulation (as these maps are implementation detail)

uui-core/src/data/processing/views/Tree.ts
```diff
@@ -223,47 +215,71 @@ export class Tree<TItem, TId> {
...
     public isFlatList() {
         return this.byParentId.size <= 1;
     }
 
-    public append(itemsToAdd: TItem[]) {
-        if (!itemsToAdd || itemsToAdd.length === 0) {
+    public patch(
+        items: TItem[],
+        isDeletedProp?: keyof TItem,
+        comparator?: (newItem: TItem, existingItem: TItem) => number,
```
jakobz, line 229: Let's provide a default value for comparator. I suggest something like () => -1 - to insert new items at the top of the list by default.

uui-core/src/data/processing/views/__tests__/Tree.patch.test.ts
```diff
@@ -0,0 +1,204 @@
...
+            [updatedItem2],
+            'isDeleted',
+            ({ id }, { id: existingId }) => (id > existingId) ? 1 : -1,
+        );
+        expect(patchedTree2['byId'].has(updatedItem2.id)).toBeTruthy();
+        expect(patchedTree2['byId'].get(updatedItem2.id)).toEqual(updatedItem2);
+        expect(patchedTree2 === tree).toBeTruthy();
+
+        expect(patchedTree2['byParentId'].get(updatedItem.parentId)).toEqual([110, 120]);
+    });
```
jakobz, line 203: Let's add test(s) to check that:
- we can insert 2 new items to the top of the list, with default comparator. Items should be inserted in the order they present in array: patch = [1,2] , tree looks like [2,1,...existing]
- we can insert 2 new items at the bottom of the list. Patch = [1,2], comparator: () => 1, tree: [...existing, 1, 2,]

uui-core/src/data/processing/views/Tree.ts
Kuznietsov: Done.

jakobz (approved):

PR #1100, opened 2023-03-09
[SlateEditor] Fix table creation

dimaanj:
## Summary

Fix table in SlateEditor:
- Adds header row on creation
- Now first cell selected after creation
- Preserves initial table width
- Now resizing works  
- Hides `remove table` button from plate lib
- Now horizontal scroll displayed only when needed (on squeezing browser window)

### DIFF
--- a/uui-editor/src/plate/plugins/tablePlugin/tablePlugin.tsx
+++ b/uui-editor/src/plate/plugins/tablePlugin/tablePlugin.tsx
@@ -218,9 +218,7 @@ const Table = (props: any) => {
     }, [element, cellPath, rowPath, cellEntries]);
 
     const colSizesRef = useRef(data.cellSizes);
-    colSizesRef.current = useTableColSizes(
-        colSizesRef.current ? { ...props.element, colSizes: colSizesRef.current } : props.element
-    );
+    colSizesRef.current = useTableColSizes({ ...props.element, colSizes: colSizesRef.current });
 
     const tableWidth = colSizesRef.current.reduce((acc: number, cur: number) => acc + cur, 0);
     return (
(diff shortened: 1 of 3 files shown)


uui-editor/src/plate/plugins/tablePlugin/tablePlugin.tsx
```diff
@@ -205,21 +209,29 @@ const Table = (props: any) => {
...
+        colSizesRef.current ? { ...props.element, colSizes: colSizesRef.current } : props.element
+    );
+
+    const tableWidth = colSizesRef.current.reduce((acc: number, cur: number) => acc + cur, 0);
     return (
-        <div className={ cx(css.wrapper, tableCSS.tableWrapper) }>
+        <div className={ cx(tableCSS.tableWrapper) }>
             <TableElement
                 { ...props }
+                styles={ { root: { width: tableWidth, "& > div": { visibility: 'hidden' } } } }
```
AlekseyManetov, line 228: can we move this styles to the scss and provide via class?

uui-editor/src/plate/plugins/tablePlugin/tablePlugin.tsx
```diff
@@ -173,7 +177,7 @@ const Table = (props: any) => {
 
     const renderToolbar = useCallback(() => {
         return (
-            <div className={ cx(css.imageToolbar, 'slate-prevent-blur') }>
+            <div className={ cx(imageBlockCss.imageToolbar, 'slate-prevent-blur') }>
```
AlekseyManetov, line 180: can we move this styles and may be wrapper also inside Toolbar component to not add them directly for tables and images?

AlekseyManetov (changes requested):

PR #1107, opened 2023-03-13
[Datasources] views subscription rework in datasources.

Kuznietsov:
### Summary

Changed a way of view's subscription/unsubscription.


Kuznietsov:
Possibly, will be revisited in the future.

PR #1114, opened 2023-03-15
[Dropdown] Change `||` to `??` when choosing `closeDelay` value with `closeOnMouseLeave` set to `boundary`

### LINKED ISSUE #1113: [Tooltip] Can't use `closeOnMouseLeave="boundary"` with `closeDelay={0}`
## Description

When using `closeOnMouseLeave="boundary"` and  `closeDelay={0}`, the closing delay is ignored, and a default fallback is used.

## Steps to Reproduce

1. Go to this [CodeSandbox](https://codesandbox.io/s/uui-tooltip-close-delay-and-close-mouse-leave-issue-xf5lwt?file=/example.tsx:247-257)
2. Hover over "Button 1", then "Button 2", then "Button 1". It is reproduced more often when hovering over paddings of the button instead of the text content

## Actual result

The tooltips stick on the screen and hide after changing the hover target after a second or two.

## Expected result

The tooltips hide immediately after changing the hover target.

## Environment

UUI version: 4.10.1
Browser name: Google Chrome v.111.0.5563.65
Operating System: Windows 10 64-bit

## Notes

Most probably, the bug is caused by this line:
https://github.com/epam/UUI/blob/main/uui-components/src/overlays/Dropdown.tsx#L166
Since `0` is a "falsy" value, the logical OR operator (`||`) chooses the fallback instead. To fix this issue, it is necessary to change it the to nullish coalescing operator (`??`), which chooses the fallback only if the value is `null` or `undefined`.

Until it is fixed, one can use `closeDelay={1}` instead.

vasilii-kovalev:
Without the change, the value `0` is ignored, which prevents dropdowns/tooltips from closing immediately.

Closes #1113.

### DIFF
--- a/uui-components/src/overlays/Dropdown.tsx
+++ b/uui-components/src/overlays/Dropdown.tsx
@@ -163,7 +163,7 @@ export class Dropdown extends React.Component<DropdownProps, DropdownState> {
                 this.handleOpenedChange(false);
             }
         } else if (this.isInteractedOutside(e) && !this.closeDropdownTimerId) { // User cursor in boundary area, but not inside toggler or body
-            this.setCloseDropdownTimer(this.props.closeDelay || 1500);
+            this.setCloseDropdownTimer(this.props.closeDelay ?? 1500);
         } else if (!this.isInteractedOutside(e) && this.closeDropdownTimerId) { // User returned to the toggler or body area, we need to clear close timer
             this.clearCloseDropdownTimer();
         }


vasilii-kovalev:
On the line 160 there is also a check like `if (this.props.closeDelay) {`. I don't know if it should be changed as well or not (to `if (this.props.closeDelay !== undefined) {`, for example).

AlekseyManetov:
> On the line 160 there is also a check like `if (this.props.closeDelay) {`. I don't know if it should be changed as well or not (to `if (this.props.closeDelay !== undefined) {`, for example).

I guess here we don't need anything to rework, because if we have `closeDelay === 0`, it's not necessary to set up a timer.
Overall it's a little bit strange to pass 0 for `closeDelay`, better just to remove this prop at all. 
But anyway your changes are also valid, thanks.

PR #1133, opened 2023-03-19
[SlateEditor] Fix table toolbar appearing in FF

dimaanj:
## Summary

Tollbar jumps to the edge of the screen on clicking empty cell.

MicrosoftTeams-image

### DIFF
--- a/uui-editor/src/implementation/Toolbar.tsx
+++ b/uui-editor/src/implementation/Toolbar.tsx
@@ -1,11 +1,12 @@
-import React, { useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import { Popper } from 'react-popper';
-import { usePlateEditorState, isEditorFocused, getSelectionBoundingClientRect } from '@udecode/plate';
+import { usePlateEditorState, isEditorFocused, getDefaultBoundingClientRect, ClientRectObject, PlateEditor, Rect } from '@udecode/plate';
 import { Portal } from '@epam/uui-components';
 import cx from "classnames";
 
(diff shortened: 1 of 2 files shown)


uui-editor/src/implementation/Toolbar.tsx
```diff
@@ -16,19 +17,33 @@ interface ToolbarProps {
...
+    if (!domSelection || domSelection.rangeCount < 1) {
+        return getDefaultBoundingClientRect();
+    }
+
+    const domRange = domSelection.getRangeAt(0);
+
+    const clientRect = domRange.getBoundingClientRect();
+    if(clientRect.bottom === 0 && clientRect.right === 0) {
+        return domSelection.anchorNode.parentElement.getBoundingClientRect();
+    }
```
dimaanj, line 32: This is the only difference between **[util](https://github.com/udecode/plate/blob/main/packages/floating/src/utils/getSelectionBoundingClientRect.ts)** provided by plate

dimaanj: At some point on clicking empty cell in the table, `window.getSelection()` starts to select `<span>` instead of `text` which leads to `getBoundingClientRect()` returns 0,0,0,0 for each of dimentions.

dimaanj: Here we selecting parent element of current selection and getting `getBoundingClientRect()` which returns valid dimentions.

dimaanj: Discussed offline

PR #1188, opened 2023-04-04
[SlateEditor] Fix from word table `copy/paste` action

dimaanj:
## Summary

This PR fixes copy/paste from word and improves attachment implementation.


uui-editor/src/plate/plugins/attachmentPlugin/AttachmentBlock.tsx
```diff
@@ -37,14 +37,14 @@ export function AttachmentBlock(props: any) {
...
-    const [fileName, setFileName] = useState(element.data.fileName || element.fileName);
+    const [name, setName] = useState(element.data.name || element.name);
 
     const changeName = (name: string) => {
         setElements(editor, {
             ...element,
             data: {
                 ...element.data,
-                fileName: name,
+                name: name,
```
AlekseyManetov, line 47: it's not safe to change the slate state object fields, because it will cause integration issues with already created states from old version. Let's leave here `fileName` field as earlier.

AlekseyManetov (changes requested):

uui-editor/src/plate/plugins/attachmentPlugin/AttachmentBlock.tsx
dimaanj: Fixed

Issue #1180, opened 2023-03-31
Pickers - 'implicit' cascade selection mode


jakobz:
## Description

The PickerInputs and data sources have a ```cascadeSelection``` option.
When ```cascadeSelection=true```, selecting a parent item automatically selects all of its children.
There is also a reverse logic, where selecting the last child of a parent automatically selects the parent.
This process is recursive. 

If the data tree is large, the user can select a lot in a single click. 
For example, in our Locations dataset, a user can select all cities of a continent in a single click. 
This leads to certain issues:

- The user gets a huge list of selected items in the Picker's body, which often doesn't fit, and we collapse it to a "N selected" message. So, after selecting a single item, the user doesn't see what's exactly selected.
- In a `LazyDataSource`, on selecting a parent, we need to load the whole sub-tree recursively, which can be time-consuming and resource-expensive.
- We often put such selection values in the URL, e.g., in case of table filters value. URLs have a limited size, and huge URLs are scary and inconvenient to use.
- We often pass such arrays of selected items to the server, and there can be performance issues if they get huge.

## Solution

In many cases, the user can understand from the context that selecting a parent implicitly means that all children are selected.
For example, consider our Locations picker, which contains the tree of continents, countries, and cities.
When the user selects "France," it is enough to show only "France" in selected items.

Other app logic can be adjusted to handle this case correctly. For example, at the server, when filtering items, instead of having just ```where cityId in [1,2,3]```, we can classify selected item types and put something like ```where countryId in [1] or cityId in [2,3,4]```.

To support this case, let's do the following:

- change 'cascadeSelection' prop type from ```boolean``` to ```boolean | 'implicit' | 'explicit'```. 
   - boolean is left for compatibility, 'true' value is the same as 'explicit' - meaning that we'll keep the old behavior
   - let's document this as following
 ```
/**
 * Controls how selection (checking items) of a parent node affects selection of its all children, and vice versa.
 * - false: All nodes are selected independently (default).
 * - true or 'explicit': Selecting a parent node explicitly selects all its children. 
 *    Unchecking the last parent's child unchecks its parent.
 * - 'implicit': Selecting a parent node means that all children are considered checked. 
 *    The user sees all these nodes as checked on the UI, but they only see the selected parent in the PickerInput tags, and only the checked parent is present in the Picker's value or DataSourceState.checked array. 
[...]

PR #1194, opened 2023-04-05
[Pickers] `implicit` cascade selection mode.

Kuznietsov:
## Summary
Closes: https://github.com/epam/UUI/issues/1180

### New behavior
https://user-images.githubusercontent.com/22456368/230328499-3343513d-6b74-415e-b8ef-a21458da03cf.mov

### Docs updates
Screenshot 2023-04-06 at 18 05 57


uui-components/src/pickers/PickerBase.tsx
```diff
@@ -169,10 +172,10 @@ export abstract class PickerBase<TItem, TId, TProps extends PickerBaseProps<TIte
     getView(): IDataSourceView<TItem, TId, any> {
         return this.props.dataSource.getView(this.getDataSourceState(), this.handleDataSourceValueChange, {
             getRowOptions: this.getRowOptions,
-            cascadeSelection: this.props.cascadeSelection,
             getSearchFields: this.props.getSearchFields || ((item: TItem) => [this.getName(item)]),
```
Kuznietsov, line 175: @jakobz, @AlekseyManetov, have no idea, if this default is specific to the current component. I'm scared to align this option with `cascadeSelection` and `sortBy` to prevent overriding the options from the datasource. On the one hand, it is required if some value is provided at the datasource and not provided to the Picker, but on the other hand, if some users haven't specified getSearchFields and rely on the current behavior, it will break their current flow. WDYT? How can we perform this required change?

AlekseyManetov: Let's assume that for cases with PickerInput this props will be passed directly to the picker, because here we need to have a default behavior. Let's return to this moment when rework our Picker Input api.

Issue #915, opened 2022-12-30
[Picker Dropdown/Dropdown Menu] Add delay for closure/open trigger of submenu, etc.


DzmitryTrubchyk:
- Add delay for closure/open trigger of submenu
- Align opened submenu by 1st item in dropdown, rather than dropdown container

[Find design & details here](https://www.figma.com/file/lXOupesvjg7g53YKmznInc/UUI-improvements?node-id=0%3A1&t=KqfmTtbgDwPEvmtP-1)
![image](https://user-images.githubusercontent.com/44213241/210064364-4c74def1-5ce8-4b07-9aa9-606507cf7634.png)

DzmitryTrubchyk:
Reviewed the task, delays are fine. Align of nested levels shpould be like on the screenshot below, if possible:
![Image](https://github.com/epam/UUI/assets/44213241/06f3ba5a-2835-4c1f-8b45-36a19dfa3ea2)

To make 1st items aligedn by on horizontal line
![Image](https://github.com/epam/UUI/assets/44213241/f94db39f-16cd-47b1-8c9a-c00e91f1a2df)


@vik753, @AlekseyManetov

NatalliaAlieva:
> Reviewed the task, delays are fine. Align of nested levels shpould be like on the screenshot below, if possible: ![Image](https://user-images.githubusercontent.com/44213241/247944549-06f3ba5a-2835-4c1f-8b45-36a19dfa3ea2.png)
> 
> To make 1st items aligedn by on horizontal line ![Image](https://user-images.githubusercontent.com/44213241/247945127-f94db39f-16cd-47b1-8c9a-c00e91f1a2df.png)
> 
> @vik753, @AlekseyManetov

This case has been moved to this [ticket](https://github.com/epam/UUI/issues/1436).

NatalliaAlieva:
Released in 5.0.0 ver.

PR #1211, opened 2023-04-12
Feature/Dropdown & DropdownMenu #915

vik753:


uui-components/src/overlays/Dropdown.tsx
```diff
@@ -153,17 +153,16 @@ export class Dropdown extends React.Component<DropdownProps, DropdownState> {
     }
 
     private handleMouseMove = (e: MouseEvent) => {
-        if (this.isInteractedOutside(e) && !this.isClientInArea(e)) { // User leave boundary area, close dropdown immediately or with this.props.closeDelay
-            this.clearCloseDropdownTimer();
+        if (this.isInteractedOutside(e) && this.isClientInArea(e) && !this.closeDropdownTimerId) { // User cursor in boundary area, but not inside toggler or body
             this.clearOpenDropdownTimer();
-
-            if (this.props.closeDelay) {
+            this.clearCloseDropdownTimer();
```
AlekseyManetov, line 158: It doesn't make sense to `clearCloseDropdownTimer` here, because we have a condition with  !this.closeDropdownTimerId in this branch

AlekseyManetov (changes requested):

Issue #1165, opened 2023-03-29
PickerInput - impossible to use TextInput / SearchInput as a toggler


jakobz:
## Steps to Reproduce

Here's sandbox: https://codesandbox.io/s/uui-picker-input-with-textinput-as-toggler-mmft9o?file=/Example.tsx

1. Type 'a' into the input
2. Dropdown opens
3. Click any city
4. Dropdown closes, and no city is not selected

Expected:
- city is selected
- dropdown shouldn't close, until I click outside 
- is we set selectionMode='single' - dropdown should close after selecting

Note that selection with keyboard still works. E.g. you can press Down and then Enter, and get items selected. I suspect we have defect in the dropdown closing logic, maybe click outside detection is broken in this case.

The most common usage for this is search input, e.g. to find a person, and redirect to some page after user selects one.
Here's an example:
https://codesandbox.io/s/uui-search-dropdown-with-navigate-on-select-02hvl5?file=/Example.tsx 

Let's add a similar example to our PickerInput docs.

NatalliaAlieva:
Released in 5.0.0 ver.

PR #1213, opened 2023-04-12
[PickerInput]: fixed behavior of PickerInput for alternative Inputs.

Kuznietsov:
### Summary
Closes: https://github.com/epam/UUI/issues/1165

### DIFF
--- a/uui-components/src/pickers/PickerInputBase.tsx
+++ b/uui-components/src/pickers/PickerInputBase.tsx
@@ -249,7 +249,6 @@ export abstract class PickerInputBase<TItem, TId, TProps> extends PickerBase<TIt
             prefix,
             suffix,
             onFocus: this.props.onFocus,
-            onBlur: this.handleBlur,
             onClear: this.handleClearSelection,
             selection: selectedRows,
             placeholder: this.getPlaceholder(),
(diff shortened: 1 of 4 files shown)


uui-components/src/pickers/PickerInputBase.tsx
```diff
@@ -315,7 +315,8 @@ export abstract class PickerInputBase<TItem, TId, TProps> extends PickerBase<TIt
         });
     }
 
-    handleBlur = (e: React.FocusEvent<HTMLElement>) => {
+    handleClickOutside = (e: React.FocusEvent<HTMLElement>) => {
```
AlekseyManetov, line 318: I suppose that this function will be just a handler for body closing, not an event handler related to any specific case like click outside or blur, so let's rename it to 'closePickerBody', remove from here `this.props.onBlur(e)` and `e` parameter and provide this blur handler directly via `getTogglerProps`.

uui-components/src/pickers/PickerInputBase.tsx
```diff
@@ -315,7 +315,8 @@ export abstract class PickerInputBase<TItem, TId, TProps> extends PickerBase<TIt
...
+        console.log('handle click outside');
```
AlekseyManetov, line 319: Remove console.log)

uui-core/src/types/props.ts
```diff
@@ -97,6 +97,7 @@ export interface IDropdownToggler extends IHasCaption, IClickable {
     isDropdown?: boolean;
     /** Called when associated dropdown should open or close  */
     toggleDropdownOpening?: (value: boolean) => void;
+    handleClickOutside?: (e: React.FocusEvent<HTMLElement>) => void;
```
AlekseyManetov, line 100: This handler will be specific for PickerInput case, in any other case dropdown toggler won't receive it, so let's move it to the `PickerTogglerProps` interface.

AlekseyManetov (changes requested):

Issue #1131, opened 2023-03-17
[PickerInput]: hover doesn't appear on "parent" when pointing the mouse


NatalliaAlieva:
## Steps to Reproduce

1. Open [Codesandbox](https://codesandbox.io/s/cascadesingleselection-forked-fi2top?file=/src/App.tsx).
2. Expand the whole tree.
3. Position the mouse over the "parent".

## Actual result

Hover doesn't appear on the "parent". Hover appears only on the "child".

https://user-images.githubusercontent.com/57751283/225926157-ef0e4a5e-d64f-4b04-ae27-c9e87b809850.mp4

## Expуcted result

The hover appears over both "parents" and "children". 

## Environment

* UUI version: 4.10.1
* Browser name: any
* Operating System: any

AlekseyManetov:
Released in 5.0.0 ver.

PR #1234, opened 2023-04-19
[PickerInput]: fixes Hover doesn't appear on "parent" when pointing the mouse.

Kuznietsov:
### Summary
Closes: https://github.com/epam/UUI/issues/1131

https://user-images.githubusercontent.com/22456368/233044571-ae8b0da8-c07c-47c9-b386-c24b3f3866e7.mov


uui-components/src/pickers/DataPickerRow.tsx
```diff
@@ -41,7 +40,7 @@ export class DataPickerRow<TItem, TId> extends React.Component<DataPickerRowProp
                 ...this.props.rawProps,
             } }
             ref={ this.rowNode }
-            cx={ [clickHandler && this.props.isFocused && uuiMod.focus, this.props.cx] }
+            cx={ [(clickHandler || this.props.isFoldable) && this.props.isFocused && uuiMod.focus, this.props.cx] }
```
AlekseyManetov, line 43: Do we have a case when we don't have a `clickHandler`, but have `this.props.isFoldable`?

AlekseyManetov (changes requested):

Issue #89, opened 2021-07-12
DataSources - handle missing items in selectedId and checked gracefully


jakobz:
DataSources receive selectedId and checked fields via DataSourceState. These fields contain item IDs, to be displayed in several scenarios - like "show selection" in Picker.

In some cases, items with these IDs wouldn't be found in data sources' data. LazyDataSource would attempt to load such items, but this can fail as well.
 
This can be caused by business logic mistakes, or by misusing DataSources. E.g. 
- no correct handling of ```ids``` param in LazyDataSource's api callback
- filtering items outside of Data Sources, making them impossible to access selected items. E.g. in dependent pickers scenario.
- items selected in past and stored in DB, currently are deleted
- items selected in past (or by another user) are not accessible to current user

These issues can be fixed by user by:
- including soft-deleted items to the list if there are selected
- filtering out such items before showing in UI 
- using datasource's built-in ```filter``` prop, instead of filtering data outside the datasource
- skipping filter when LazyDataSource is queried items by id (with ```ids``` api request field)

However, such issues are hard to spot. We should handle them gracefully, and help projects to spot them.

Kuznietsov:
@jakobz, there is another way of catching such an error: selecting some item and then, removing it from the DB.
And here is a question, are we going to handle it in a safe way on the side of UUI? Thanks for the answer.


cc @AlekseyManetov

jakobz:
Let's do following:

If there is an item in selectedId or checkedIds, which is 
a) can't be found along items passed to ArrayDataSource
b) haven't returned via api call with ids param in LazyDataSource

then:
- console.warn the following message
"DataSource can't find selected/checked items with following IDs: 43, 21, 5. Read more here: https://github.com/epam/UUI/issues/89."
- we return this TItem as undefined via all our APIs (rows, getById, etc.)
- our internal logic doesn't crash and works as usual with such items
- our default render logic in Pickers renders "Unknown" for names of such items (in picker's tags, in PickerList items, in PickerRows when 'show selected' is turned on
- wrap user-passed getName with try/catch, and render "Unknown" if error occured

AlekseyManetov:
released in 5.0.0

PR #1275, opened 2023-04-27
[Datasources]: not found records handling

Kuznietsov:
### Summary

Closes: https://github.com/epam/UUI/issues/89


uui/components/filters/FiltersPanelItem.tsx
```diff
@@ -153,6 +155,18 @@ function FiltersToolbarItemImpl(props: FiltersToolbarItemProps) {
         return predicate ? props.value?.[predicate] : props.value;
     };
 
+    const getName = (item: DataRowProps<any, any>, footerProps: PickerFilterConfig<any>) => {
```
AlekseyManetov, line 158: Do we need here second param, can we use props from closure? Also we get name only for pickers, let's name it more specific, for example - getPickerItemName()

AlekseyManetov (changes requested):

app/src/docs/_examples/pickerInput/LazyTreeInput.example.tsx
```diff
@@ -9,7 +9,7 @@ import { Location } from '@epam/uui-docs';
 
 export default function LazyTreePicker() {
     const svc = useUuiContext();
-    const [value, onValueChange] = useState<string[]>();
```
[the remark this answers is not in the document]
Kuznietsov: It was added for the purpose of testing `unknown in PickerInput` and `show only selected` scenarios during the review process.

uui/components/filters/FiltersPanelItem.tsx
Kuznietsov: Done. But we can't take props from closure, because `switch` in this case is playing the role of the guard, which determines the exact type of props from the union of types.

Issue #1270, opened 2023-04-26
Update color palete for Loveship and Promo


AlekseyManetov:
## To Do
Update color palete according to the design specs - https://www.figma.com/file/XvNqYvKrGf8PP1wCedPVQu/UUI-Components-Main-Branch-(Sandbox)?node-id=2123-471275&t=UqeP3Ek9DjTAdIEc-0

Update color variables in themes, @epam/assets package and ./assets folder inside skins.

Also this change will affect some components color props:
- In `Button` for "night500" and "gray50", mark them as deprecated and bind them in styles to "night600/gray60". Add a new "gray" value that refers to "gray60". In property explorer leave only 'gray' value.
- In `Tooltip` for "night900" and "gray90", mark them as deprecated and bind them in styles to "night800/gray80". Add a new "gray" value that refers to "gray80". In property explorer leave only 'gray' value.

NatalliaAlieva:
Released in 5.0.0 ver.

PR #1295, opened 2023-05-03
Feature/update color palete #1270

vik753:


app/src/docs/_props/epam-promo/components/buttons/button.props.tsx
```diff
@@ -29,7 +26,7 @@ const ButtonDoc = new DocBuilder<ButtonProps & ButtonMods>({ name: 'Button', com
     ])
     .prop('color', {
         renderEditor: (editable, examples) => <ColorPicker colors={ examples.map((i) => ({ value: i, hex: colors[i] })) } { ...editable } />,
-        examples: allButtonColors,
+        examples: allButtonColors as any,
```
AlekseyManetov, line 29: why do we need any here?

AlekseyManetov (changes requested):

app/src/docs/_props/epam-promo/components/buttons/button.props.tsx
vik753: fixed

PR #1366, opened 2023-05-24
[SlateEditor] Fix splitter remove

dimaanj:
## Summary

Now when slecting all content using command + a -> splitter will be removed


dimaanj:
Closing it. Decided to create issue on plate editor.

PR #1386, opened 2023-05-31
[PickerInput]: tests.

Kuznietsov:
### Summary
Added tests for PickerInput.

### DIFF
--- a/uui-components/src/pickers/PickerToggler.tsx
+++ b/uui-components/src/pickers/PickerToggler.tsx
@@ -118,6 +118,7 @@ function PickerTogglerComponent<TItem, TId>(props: PickerTogglerProps<TItem, TId
                 ref={ inputContainer }
                 aria-haspopup={ true }
                 autoComplete="no"
+                data-testid="uui-PickerToggler-input"
                 aria-required={ props.isRequired }
                 aria-disabled={ props.isDisabled }
                 aria-readonly={ props.isReadonly }
--- a/uui-components/src/pickers/PickerToggler.tsx
+++ b/uui-components/src/pickers/PickerToggler.tsx
@@ -154,7 +154,13 @@ function PickerTogglerComponent<TItem, TId>(props: PickerTogglerProps<TItem, TId
         [props.isOpen, props.closePickerBody],
     );
 
-    const icon = props.icon && <IconContainer icon={ props.icon } onClick={ props.onIconClick } />;
+    const icon = props.icon && (
+        <IconContainer
+            icon={ props.icon }
+            onClick={ props.onIconClick }
+            rawProps={ { 'data-testid': 'uui-PickerToggler-iconContainer' } }
+        />
+    );
 
     return (
         <div
--- a/uui-components/src/pickers/PickerToggler.tsx
+++ b/uui-components/src/pickers/PickerToggler.tsx
@@ -190,7 +190,7 @@ function PickerTogglerComponent<TItem, TId>(props: PickerTogglerProps<TItem, TId
                             icon={ props.cancelIcon }
                             tabIndex={ -1 }
                             onClick={ handleCrossIconClick }
-                            rawProps={ { role: 'button' } }
+                            rawProps={ { role: 'button', 'aria-label': 'Clear' } }
                         />
                     )}
                     {props.isDropdown && <IconContainer icon={ props.dropdownIcon } flipY={ props.isOpen } cx="uui-icon-dropdown" onClick={ closeOpenedPickerBody } />}
--- a/uui-components/src/pickers/PickerToggler.tsx
+++ b/uui-components/src/pickers/PickerToggler.tsx
@@ -182,7 +182,10 @@ function PickerTogglerComponent<TItem, TId>(props: PickerTogglerProps<TItem, TId
             onKeyDown={ props.onKeyDown }
             { ...props.rawProps }
         >
-            <div className={ cx(css.body, !props.isSingleLine && props.pickerMode !== 'single' && css.multiline) }>
+            <div
+                className={ cx(css.body, !props.isSingleLine && props.pickerMode !== 'single' && css.multiline) }
+                data-testid="uui-PickerToggler-body"
+            >
                 {props.iconPosition !== 'right' && icon}
                 {props.pickerMode !== 'single' && renderItems()}
                 {renderInput()}
--- a/uui-components/src/pickers/PickerToggler.tsx
+++ b/uui-components/src/pickers/PickerToggler.tsx
@@ -97,7 +97,15 @@ function PickerTogglerComponent<TItem, TId>(props: PickerTogglerProps<TItem, TId
 
     const renderInput = () => {
         const isSinglePickerSelected = props.pickerMode === 'single' && props.selection && !!props.selection[0];
-        const placeholder = isSinglePickerSelected ? props.getName(props.selection[0]?.value) : props.placeholder;
+        let placeholder: string;
+        if (!isSinglePickerSelected) {
+            placeholder = props.placeholder;
+        }
+         
+        if (isSinglePickerSelected) {
+            placeholder = props.selection[0].isLoading ? undefined : props.getName(props.selection[0]?.value);
+        }
+    
         const value = props.disableSearch ? null : props.value;
         if (props.searchPosition !== 'input' && props.pickerMode === 'multi' && props.selectedRowsCount > 0) {
             return null;
--- a/uui/components/pickers/PickerInput.tsx
+++ b/uui/components/pickers/PickerInput.tsx
@@ -99,7 +99,7 @@ export class PickerInput<TItem, TId> extends PickerInputBase<TItem, TId, PickerI
             <IEditableDebouncer
                 value={ targetProps.value }
                 onValueChange={ this.handleTogglerSearchChange }
-                render={ (editableProps) => renderTarget({ ...this.getTogglerMods(), ...targetProps, ...editableProps }) }
+                render={ (editableProps) => renderTarget({ ...this.getTogglerMods(), ...targetProps, ...editableProps, rawProps: { ...targetProps.rawProps, 'data-testid': 'uui-PickerInput-target' } }) }
             />
         );
     }
--- a/uui/components/pickers/PickerInput.tsx
+++ b/uui/components/pickers/PickerInput.tsx
@@ -62,6 +62,10 @@ export class PickerInput<TItem, TId> extends PickerInputBase<TItem, TId, PickerI
         ) : (
             <DataPickerRow
                 { ...rowProps }
+                rawProps={ {
+                    ...rowProps.rawProps,
+                    'data-testid': `uui-PickerInput-item-${rowProps.rowKey}`,
+                } }
                 key={ rowProps.rowKey }
                 borderBottom="none"
                 size={ this.getRowSize() }
--- a/uui/components/pickers/PickerInput.tsx
+++ b/uui/components/pickers/PickerInput.tsx
@@ -99,7 +99,15 @@ export class PickerInput<TItem, TId> extends PickerInputBase<TItem, TId, PickerI
             <IEditableDebouncer
                 value={ targetProps.value }
                 onValueChange={ this.handleTogglerSearchChange }
-                render={ (editableProps) => renderTarget({ ...this.getTogglerMods(), ...targetProps, ...editableProps, rawProps: { ...targetProps.rawProps, 'data-testid': 'uui-PickerInput-target' } }) }
+                render={ (editableProps) => renderTarget({
+                    ...this.getTogglerMods(),
+                    ...targetProps,
+                    ...editableProps,
+                    rawProps: {
+                        ...targetProps.rawProps,
+                        'data-testid': 'uui-PickerInput-target',
+                    },
+                }) }
             />
         );
     }
--- a/uui/components/pickers/PickerToggler.tsx
+++ b/uui/components/pickers/PickerToggler.tsx
@@ -53,6 +53,9 @@ function PickerTogglerComponent<TItem extends string, TId>(props: PickerTogglerP
 
     const renderItem = (row: DataRowProps<TItem, TId>) => (
         <Tag
+            rawProps={ {
+                'data-testid': `uui-PickerToggler-item-${row.rowKey}`,
+            } }
             key={ row.rowKey }
             caption={ getCaption(row) }
             tabIndex={ -1 }
--- a/uui/components/pickers/__tests__/PickerInput.test.tsx
+++ b/uui/components/pickers/__tests__/PickerInput.test.tsx
@@ -227,14 +227,20 @@ describe('PickerInput', () => {
             getName: ({ name }) => name,
         });
 
+        const selectedItemsNames1 = screen.queryAllByRole('button')
+            .filter((button) => button.getAttribute('aria-label') !== 'Clear')
+            .map((button) => button.textContent?.trim());
+        
+        expect(selectedItemsNames1).toEqual(['', '']);
+
         expect(dom.input?.getAttribute('placeholder')?.trim()).toEqual('Please select');
    
         await delayAct(100);
 
-        const selectedItemsNames = screen.queryAllByRole('button')
+        const selectedItemsNames2 = screen.queryAllByRole('button')
             .filter((button) => button.getAttribute('aria-label') !== 'Clear')
-            .map((button) => button.textContent);
+            .map((button) => button.textContent?.trim());
         
-        expect(selectedItemsNames).toEqual(['Elementary+', 'Pre-Intermediate']);
+        expect(selectedItemsNames2).toEqual(['Elementary+', 'Pre-Intermediate']);
     });
 });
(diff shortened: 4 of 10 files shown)


uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+            const { mocks, dom } = await setupPickerInputForTest({
+                value: undefined,
+                getName: ({ name }) => name,
+                selectionMode: 'multi',
+                cascadeSelection: false,
+                dataSource: mockTreeLikeDataSourceAsync,
+            });
+    
+            fireEvent.click(dom.input as HTMLElement);
+            await delayAct(100);
```
siarheiyelin, line 430: please avoid such kind of delays. use findBy* queries instead (https://testing-library.com/docs/guide-disappearance/)

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+            const { mocks, dom } = await setupPickerInputForTest({
+                value: undefined,
+                getName: ({ name }) => name,
+                selectionMode: 'multi',
+                cascadeSelection,
+                dataSource: mockTreeLikeDataSourceAsync,
+            });
+    
+            fireEvent.click(dom.input as HTMLElement);
+            await delayAct(100);
```
siarheiyelin, line 457: please avoid such kind of delays.

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+        ('should pick multiple elements with cascadeSelection = %s', async (cascadeSelection) => {
+            const { mocks, dom } = await setupPickerInputForTest({
+                value: undefined,
+                getName: ({ name }) => name,
+                selectionMode: 'multi',
+                cascadeSelection,
+                dataSource: mockTreeLikeDataSourceAsync,
+            });
+    
+            fireEvent.click(dom.input as HTMLElement);
```
siarheiyelin, line 456: it's better to cast this input to HTMLElement in the "setup" function instead.

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
         expect(tree).toMatchSnapshot();
     });
+    
+    describe('[selectionMode single]', () => {
+        it('should select & clear option', async () => {
+            const { dom, mocks } = await setupPickerInputForTest({
+                value: undefined,
+                selectionMode: 'single',
+            });
+            expect(dom.input?.getAttribute('placeholder')?.trim()).toEqual('Please select');
```
siarheiyelin, line 171: please avoid any DOM specific API as much as possible. In this case it's better to check it as: ```expect(screen.queryByPlaceholderText('Please select', { exact: false })).toBeInTheDocument();```

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+            fireEvent.click(dom.input as HTMLElement);
```
siarheiyelin, line 172: type cast should be done only once in the setup function

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+        it('should render names of items by getName', async () => {
+            const { mocks, dom } = await setupPickerInputForTest<TestItemType, number>({
+                value: 3,
+                selectionMode: 'single',
+                getName: ({ name }) => name,
+            });
+
+            expect(dom.input?.getAttribute('placeholder')?.trim()).toBeUndefined();
+            
+            await delayAct(100);
```
siarheiyelin, line 194: delay is redundant, use findBy* query instead

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+        
+        it('should ignore plural entity name in placeholder', async () => {
+            const { dom } = await setupPickerInputForTest({
+                value: undefined,
+                selectionMode: 'single',
+                entityName: 'Language Level',
+                entityPluralName: 'Multiple Language Levels',
+            });
+            
+            expect(dom.input?.getAttribute('placeholder')?.trim()).toEqual('Please select Language Level');
```
siarheiyelin, line 223: ```expect(screen.queryByPlaceholderText('....test here....', { exact: false })).toBeInTheDocument();```

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+            const { mocks, dom } = await setupPickerInputForTest({
+                value: undefined,
+                getName: ({ name }) => name,
+                selectionMode: 'single',
+                cascadeSelection,
+                dataSource: mockTreeLikeDataSourceAsync,
+            });
+    
+            fireEvent.click(dom.input as HTMLElement);
+            await delayAct(100);
```
siarheiyelin, line 239: delay is not needed here

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+            value: undefined,
+            minCharsToSearch: 1,
+        });
+
+        fireEvent.click(dom.input as Element);
+
+        expect(screen.queryByRole('dialog')).not.toBeInTheDocument();
+        
+        act(() => {
+            jest.useFakeTimers();
```
siarheiyelin, line 821: maybe we could use findBy* queries to wait for apperance?

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+
+        const target = screen.getByTestId('uui-PickerInput-target');
+        expect(target).toBeInTheDocument();
+        expect(target.getAttribute('type')).toBe('button');
+        
+        fireEvent.click(target);
+        
         expect(screen.getByRole('dialog')).toBeInTheDocument();
+
+        await delayAct(100);
```
siarheiyelin, line 955: it seems redundant

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+                    key={ props.rowKey }
+                    alignActions="center"
+                    renderItem={ (item, rowProps) => <PickerItem { ...rowProps } title={ item.name } /> }
+                />
+            ),
+        });
+
+        fireEvent.click(dom.input as Element);
+
+        await delayAct(100);
```
siarheiyelin, line 1070: findBy* queries can be used instead

uui/components/pickers/PickerToggler.tsx
```diff
@@ -53,6 +53,9 @@ function PickerTogglerComponent<TItem extends string, TId>(props: PickerTogglerP
 
     const renderItem = (row: DataRowProps<TItem, TId>) => (
         <Tag
+            rawProps={ {
+                'data-testid': `uui-PickerToggler-item-${row.rowKey}`,
```
siarheiyelin, line 57: as far as I remember we agreed to use semantic markup for tags, i.e. add aria-role to the wrapper div and to each Tag item. Adding "data-testid" is less preferrable.

uui/components/pickers/PickerInput.tsx
```diff
@@ -62,6 +62,10 @@ export class PickerInput<TItem, TId> extends PickerInputBase<TItem, TId, PickerI
         ) : (
             <DataPickerRow
                 { ...rowProps }
+                rawProps={ {
+                    ...rowProps.rawProps,
+                    'data-testid': `uui-PickerInput-item-${rowProps.rowKey}`,
```
siarheiyelin, line 67: as far as I remember we agreed to use semantic markup for tags, i.e. add aria-role to the wrapper div and to each Tag item. Adding "data-testid" is less preferrable.

uui-components/src/pickers/PickerToggler.tsx
```diff
@@ -144,7 +153,13 @@ function PickerTogglerComponent<TItem, TId>(props: PickerTogglerProps<TItem, TId
         [props.isOpen, props.closePickerBody],
     );
 
-    const icon = props.icon && <IconContainer icon={ props.icon } onClick={ props.onIconClick } />;
+    const icon = props.icon && (
+        <IconContainer
+            icon={ props.icon }
+            onClick={ props.onIconClick }
+            rawProps={ { 'data-testid': 'uui-PickerToggler-iconContainer' } }
```
siarheiyelin, line 160: is it possible to use only aria-label & aria-role=button?

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+            });
+    
+            const selectedItemsNames1 = screen.queryAllByTestId(/uui-PickerToggler-item/)
+                .map((button) => button.textContent?.trim());
+            
+            expect(selectedItemsNames1).toEqual(['', '']);
+    
+            expect(dom.input?.getAttribute('placeholder')?.trim()).toEqual('Please select');
+       
+            await delayAct(100);
```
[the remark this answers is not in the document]
Kuznietsov: @siarheiyelin, very nice suggestion. Thanks 👍

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+                value: 2,
+                selectionMode: 'single',
+                disableClear: false,
+            });
+
+            await delayAct(100);
+
+            const target = screen.getByTestId('uui-PickerInput-target');
+    
+            const [clearButton] = within(target).getAllByRole('button').filter((el) => el.getAttribute('aria-label') === 'Clear');
```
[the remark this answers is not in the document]
Kuznietsov: I think, that tests should be as simple, as possible. It is not really good idea to go to the definition of some check/query function. The definition of the function is not so huge to pay such a price.

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -121,51 +161,921 @@ describe('PickerInput', () => {
...
+            fireEvent.click(clearButton2);
+
+            expect(dom.input?.getAttribute('placeholder')?.trim()).toEqual('Please select');
+
+            const dialog3 = screen.getByRole('dialog');
+
+            const [clearButton3] = within(dialog3).getAllByRole('button')
+                .filter((el) => el.getAttribute('aria-label') === 'CLEAR');
+            
+            expect(clearButton3.getAttribute('aria-disabled')).toBe('true');
```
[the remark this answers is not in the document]
Kuznietsov: Unfortunately, it is not disabled. `disabled` attribute is not set. So, we can't use that function.

uui/components/pickers/__tests__/PickerInput.test.tsx
Kuznietsov: Good point, thanks.

uui/components/tables/DataTableCell.tsx
```diff
@@ -33,6 +33,10 @@ function DataTableRowAddons<TItem, TId, TCellValue>(props: DataTableCellProps<TI
                 <div key="fold" className={ css.indent } style={ { marginLeft: (row.indent - 1) * 24 } }>
                     {row.isFoldable && (
                         <IconContainer
+                            rawProps={ {
+                                'aria-label': 'Fold/Unfold',
+                                'data-testid': `uui-DataTableRowAddons-folding-arrow-${row.rowKey}`,
```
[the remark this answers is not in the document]
Kuznietsov: Fold/unfold can be in different levels, so, adding `data-testid` will make it possible to query not all the folding arrows, but the exact one. And this pattern of querying something by role and aria-label is kind of asking if somebody knows the person with long hair and blue eyes, and how is working as an engineer, instead of asking if somebody knows the person with the exact name. WDYT?

uui/components/pickers/PickerToggler.tsx
Kuznietsov: See the comment before. A see it as an antipattern. And we agreed on adding data-testids for the cases, where multiple items can be added.

uui-components/src/pickers/PickerToggler.tsx
Kuznietsov: How will you do it and why should you do that? This icon is custom, can be ignored by the user, and can be added to the picker input with some purpose (unknown). Why should we look for an icon in the array of other icons with some `aria-label`, which is not correspond to the real purpose of the button just for testing needs?

uui-components/src/pickers/PickerToggler.tsx
```diff
@@ -109,6 +117,7 @@ function PickerTogglerComponent<TItem, TId>(props: PickerTogglerProps<TItem, TId
                 ref={ inputContainer }
                 aria-haspopup={ true }
                 autoComplete="no"
+                data-testid="uui-PickerToggler-input"
```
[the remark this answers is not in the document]
Kuznietsov: How and why? If a possibility of changing the input is present, we need to know that an exact component is rendered.

uui/components/pickers/__tests__/PickerInput.test.tsx
Kuznietsov: I want to query placeholder of the exact input, not of some at the component. Why should I query over all the page (even if all the page is PickerInput component)? What if there will be inputs at the rendered rows? This example is highlighting all the problems of such an approach. It is very unstable.

uui/components/tables/DataTableCell.tsx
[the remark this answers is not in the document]
siarheiyelin: The difference is what is tested: "implementation details" (data attributes) vs "end result" (can be visible or perceivable by end user). So that test fails only when visible behavior changes, and doesn't fail if only internal details change. Same is recommended by the testing library docs https://testing-library.com/docs/queries/about/#priority

I don't know how to find this button at a specific nesting level. Maybe query by item label first, and then query by Fold/Unfold label in same container. If it's impossible or really difficult, then it's ok to leave approach with data-testid

uui/components/pickers/PickerInput.tsx
```diff
@@ -95,7 +101,15 @@ export class PickerInput<TItem, TId> extends PickerInputBase<TItem, TId, PickerI
...
                 value={ targetProps.value }
                 onValueChange={ this.handleTogglerSearchChange }
-                render={ (editableProps) => renderTarget({ ...this.getTogglerMods(), ...targetProps, ...editableProps }) }
+                render={ (editableProps) => renderTarget({
+                    ...this.getTogglerMods(),
+                    ...targetProps,
+                    ...editableProps,
+                    rawProps: {
+                        ...targetProps.rawProps,
+                        'data-testid': 'uui-PickerInput-target',
```
siarheiyelin, line 110: Here it should be possible to use this.props.rawProps?.input to pass any custom data-testid, from test itself

uui/components/pickers/PickerToggler.tsx
siarheiyelin: In this specific case we have an opportunity to improve a11y by using semantic markup (aria roles for list & items https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listitem_role)

uui-components/src/pickers/PickerToggler.tsx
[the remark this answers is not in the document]
siarheiyelin: I guess only 1 input is rendered here. probably some other inputs in the dialog, but it's another container.

uui/components/pickers/__tests__/PickerInput.test.tsx
```diff
@@ -123,55 +268,649 @@ describe('PickerInput', () => {
...
+            const clearButton = within(dom.container).getByRole('button', { name: 'Clear' });
+            expect(clearButton).toBeInTheDocument();
+            fireEvent.click(clearButton);
+            expect(PickerInputObject.getPlaceholderText(dom.input)).toEqual('Please select');
+            setProps({ disableClear: true, value: 2 });
+            expect(within(dom.container).queryByRole('button', { name: 'Clear' })).not.toBeInTheDocument();
+            expect(PickerInputObject.getPlaceholderText(dom.input)).toEqual('A1');
+        });
+
+        it('should select all', async () => {
```
AlekseyManetov, line 387: in test name 'should select all' but test check 'Clear' behaviour

AlekseyManetov (changes requested):
It would be nice to also test valueType='entity', search positions and search behavior, fold/unfold logic and 'Show only selected' button behavior

PR #1408, opened 2023-06-12
[Pickers]: moved to functional components & hooks.

Kuznietsov:
### Summary
In progress...


uui-components/src/pickers/hooks/usePicker.ts
```diff
@@ -0,0 +1,196 @@
...
+        cascadeSelection,
+    } = props;
+
+    const propsRef = useRef(props);
+    propsRef.current = props;
+
+    const pickerStateRef = useRef(pickerState);
+    pickerStateRef.current = pickerState;
+
+    const handleDataSourceValueChangeRef = useRef((newDataSourceState: DataSourceState) => {
```
Kuznietsov, line 35: @jakobz, @AlekseyManetov, please, have a look at this piece of code. It is required to create this function via ref to keep the link because a view is created and retrieved via a link to this function later.

uui/components/pickers/PickerInput.fn.tsx
```diff
@@ -0,0 +1,186 @@
...
+import { DataPickerFooter } from './DataPickerFooter';
+import { PickerItem } from './PickerItem';
+
+import css from './PickerInput.module.scss';
+
+const pickerHeight = 300;
+const pickerWidth = 360;
+
+function PickerInput<TItem, TId>(props: PickerInputBaseProps<TItem, TId> & SizeMod & IHasEditMode) {
+    const toggleModalOpening = () => {
```
Kuznietsov, line 22: @jakobz, @AlekseyManetov, this function is passed to the `usePickerInput` hook, and functions from the output of `usePickerInput`  are used inside this function. Pretty weird, but the only option I found.

Issue #1411, opened 2023-06-14
[Rating]: rating is displayed as 0.5, with the initial value equal to 0 in steps of 0.5


NatalliaAlieva:
## Steps to Reproduce
1. Open [Codesandbox](https://codesandbox.io/s/uui-forked-gd3458?file=/Example.tsx).
2. Verify how many stars are displayed by default

## Actual result
A value of 0 displays half of the star.
![value=0 step=0 5 shown as 0 5](https://github.com/epam/UUI/assets/57751283/05466ff1-379e-4e08-8790-da212b41af43)

## Expected result
With a value of 0, all stars are not fiiled.

## Environment
* UUI version: 5.0.0
* Browser name: any
* Operating System: any

Djarrieta:
Hello! Can you please assign this one to me? I am already sending the PR.

Djarrieta:
This is the related [PR](/epam/UUI/pull/1452).
There is also a [demo video](https://www.youtube.com/watch?v=SeOdyf72ljo)

NatalliaAlieva:
Hello, @Djarrieta! 
Could you please add in the Property Explore value=0?

Djarrieta:
Yes @NatalliaAlieva , I can do that. I am going to do it in a new PR. I let you know when it is send.

NatalliaAlieva:
> Yes @NatalliaAlieva , I can do that. I am going to do it in a new PR. I let you know when it is send.

Thank you!

NatalliaAlieva:
Hi @Djarrieta ,

During testing your fix, I found another one:

https://github.com/epam/UUI/assets/57751283/a606daf4-a23a-4737-a385-7e1659641572

Could you please see this case too?

Djarrieta:
It is the same issue solved in [the first PR](https://github.com/epam/UUI/pull/1452) , it has no been merge in the main branch yet, at least in production. As soon as it is merged, it is not going to be a problem any more.

AlekseyManetov:
I've cherry picked this commit to the develop branch. @NatalliaAlieva, please check this case again. @Djarrieta, please create PR to the 'develop' branch next time.

Djarrieta:
I got it, I was following the ["Fork-and-Pull" Git workflow](https://github.com/susam/gitpr) as the docs says. Next time will be Develop branch, no problem.

AlekseyManetov:
[...]

PR #1452, opened 2023-06-29
bugfix/Rating-initial-value

Djarrieta:
#### Issue link(if exists): 
https://github.com/epam/UUI/issues/1411

### Description:
The problem lies in the "checkRating" function, where it replaces the "rating" value with the "from" value. When the "step" value is undefined (which is the case most of the time), there are no issues. However, when the "step" value is set to 0.5, a problem arises. This issue can be resolved by excluding the equality in the "checkRating" function.

### Demo video:
[Demo video](https://www.youtube.com/watch?v=SeOdyf72ljo)


AlekseyManetov:
@Djarrieta Thanks for your contribution and very detailed description with video.

PR #1459, opened 2023-07-04
[Dropdown] Fix `renderTarget` props memoization

vasilii-kovalev:
### Description

Some properties (like `isInteractedOutside`), that are passed to `renderTarget` function, are created anew on each render, breaking user memoized components.

This pull request move the properties' getters to separate component's methods:
* `isInteractedOutside` (method `getIsInteractedOutside`)
* `onClick` (method `getTargetClickHandler`). Actually, the current version is safe enough, but I decided to move picking the handler to a separate method to make the code a bit cleaner

Additionally, I removed unnecessary condition part `!openOnClick` inside `getTargetClickHandler`, because it is always `true`:

```diff
if (
  openOnClick
+ || !openOnHover
- || (
-   !openOnClick
-   && !openOnHover
- )
) {
  return this.handleTargetClick;
}

return undefined;
```

### Further improvements

There is also `ref` property, but I left it untouched, because right now it is impossible to make it the component's property because its reference to `targetProps` object, returned by `Reference` component from “react-popper” library and updated on each render.

So the problem with memoization user's components is still here because of `ref`, but now it is not necessary to worry about other properties and spend time on fixing them.

### DIFF
--- a/uui-components/src/overlays/Dropdown.tsx
+++ b/uui-components/src/overlays/Dropdown.tsx
@@ -183,9 +183,23 @@ export class Dropdown extends React.Component<DropdownProps, DropdownState> {
         else this.handleOpenedChange(false);
     };
 
-    private renderTarget(targetProps: ReferenceChildrenProps) {
+    private getTargetClickHandler = () => {
         const { openOnClick, openOnHover } = this.props;
-        const handleTargetClick = (openOnClick || (!openOnClick && !openOnHover)) ? this.handleTargetClick : undefined;
+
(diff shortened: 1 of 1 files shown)


siarheiyelin:
Hello [vasilii-kovalev](https://github.com/vasilii-kovalev) could you please pull the latest develop to your branch. It should fix the issue with github actions check.

vasilii-kovalev:
I made the following updates in the pull request:
1. Built the changes on top of `develop` branch instead of `main`
2. Removed changes in the documentation example. Now only necessary changes in the component's code are presented
3. Changed target branch to `develop` instead of `main`
4. Updated title and description

Issue #1466, opened 2023-07-05
[Date Picker]: improvement according to the design specs


OlegKryvokobylskyi:
### Now
Max size=48.
![image](https://github.com/epam/UUI/assets/129292145/640109eb-2389-4229-b7d4-426b299fd8f8)

### To Do
48 size should be deprecated.
Max size should be 42.

Link to [Loveship (UUI3)](https://www.figma.com/file/M5Njgc6SQJ3TPUccp5XHQx/Loveship-aka-UUI3-(Components)?type=design&node-id=8662-290368&mode=design&t=9GljFIfcSi4OQ2bc-0)
Link to [Promo (UUI4)](https://www.figma.com/file/UyChXPLmyv5zMrOU37KdUL/Promo-aka-UUI4-(Components)?type=design&node-id=14836-271788&mode=design&t=TcfLKYaXs21SRJBA-0)

### To Do
Make footer examples for Documentation

See design:
Link to [Loveship (UUI3)](https://www.figma.com/file/M5Njgc6SQJ3TPUccp5XHQx/Loveship-aka-UUI3-(Components)?type=design&node-id=269-84908&mode=design&t=dGMyNOCShwaBwA6j-0)
Link to [Promo (UUI4)](https://www.figma.com/file/UyChXPLmyv5zMrOU37KdUL/Promo-aka-UUI4-(Components)?type=design&node-id=269%3A84908&mode=design&t=ACg8l3ad9o0289vV-1)

Default and active state should not change in width

### To Do
fix footer in calendar according to [UI Difference](https://www.figma.com/file/M41fdci2Y1ErJ17NrJxLDp/Design-Review?type=whiteboard&node-id=3%3A202&t=KyKzJkx4ytpdunli-1)

![image](https://github.com/epam/UUI/assets/129292145/435d46f0-1814-4fc1-8481-f6dc8b223b0e)

Make a basic footer for the property explorer

vik753:
https://github.com/epam/UUI/pull/1481

DzmitryTrubchyk:
Same issue as with range date picker, it doesn't open. I see Something went wrong screen and can't review @vik753

NatalliaAlieva:
> Same issue as with range date picker, it doesn't open. I see Something went wrong screen and can't review @vik753

Fixed on dev.

NatalliaAlieva:
Released in 5.1.1 ver.

PR #1481, opened 2023-07-10
Bugfix/DatePicker improvement by design specs #1466

vik753:


vik753:
Issue: https://github.com/epam/UUI/issues/1466

PR #1516, opened 2023-07-20
Feature/filters panel selected elements#1480

vik753:

### DIFF
--- a/app/src/docs/_examples/tables/FiltersPanelBasic.example.tsx
+++ b/app/src/docs/_examples/tables/FiltersPanelBasic.example.tsx
@@ -82,6 +82,7 @@ export default function FiltersPanelExample() {
                 columnKey: 'jobTitle',
                 title: 'Title',
                 type: 'multiPicker',
+                maxCount: 2,
                 dataSource: new LazyDataSource({ api: api.demo.jobTitles }),
             }, {
                 field: 'salary',
--- a/uui-core/src/types/tables.ts
+++ b/uui-core/src/types/tables.ts
@@ -225,6 +225,8 @@ type FilterConfigBase<TFilter> = {
     columnKey: string;
     isAlwaysVisible?: boolean;
     predicates?: IFilterPredicate[];
+    /** default value: 2. count of words to show in the Filter PickerToggler */
+    maxCount?: number;
 };
 
 export type PickerFilterConfig<TFilter> = FilterConfigBase<TFilter> & {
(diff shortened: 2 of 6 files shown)


app/src/docs/_examples/tables/FiltersPanelBasic.example.tsx
```diff
@@ -82,6 +82,7 @@ export default function FiltersPanelExample() {
                 columnKey: 'jobTitle',
                 title: 'Title',
                 type: 'multiPicker',
+                maxCount: 2,
```
AlekseyManetov, line 85: it's no sense to pass maxCount: 2, because 2 is a default value. Better lets add maxCount: 3, to the 'profileStatus' filter, because in this case we have short names and only 3 options.

uui-core/src/types/tables.ts
```diff
@@ -225,6 +225,8 @@ type FilterConfigBase<TFilter> = {
     columnKey: string;
     isAlwaysVisible?: boolean;
     predicates?: IFilterPredicate[];
+    /** default value: 2. count of words to show in the Filter PickerToggler */
```
AlekseyManetov, line 228: Move info about default to the end of the description:
`/** Count of words to show in the Filter toggler. By default 2 item will be shown. */`

AlekseyManetov (changes requested):

uui-core/src/types/tables.ts
vik753: fixed

uui/components/filters/FilterPanelItemToggler.tsx
```diff
@@ -29,6 +27,13 @@ export const FilterPanelItemToggler = React.forwardRef<HTMLDivElement, FilterToo
 
     const getTitle = props.predicateName ? `${props.title} ${props.predicateName}` : `${props.title}${props.selection ? ':' : ''}`;
 
+    const getSelectionText = () => props.selection.map((i, index) => (
+        <>
+            <Text key={ index } color="brand" size={ props.size } cx={ css.selection }>{i}</Text>
```
9vit917, line 32: You should use key prop in wrapper element for list Item (use some div instead Fragment)
Also it is a bad practice to use index as key

9vit917: or just use <Fragment key={...}

vik753: Thank you @9vit917  for your message. I corrected it there:
https://github.com/epam/UUI/commit/5fd1e18da4d6f11a64b0cad0fc74c6da862ac3e3

PR #1522, opened 2023-07-24
[Tag] move count from left to right according to design

DIWAKARKASHYAP:


AlekseyManetov:
This improvement required much more changes according to the design specs: various colours, outline styles, counter in component color. I guess it's not an 'good first issue', because it's required a lot of consultations and diving into details. Let's better choose another issue.

PR #1535, opened 2023-07-27
[Fix] Use LinkButton instead of Avatar for dropdown menu example

maryia-lapata:
### Description:

It was decided that `Avatar` shouldn't be a clickable element. So in scope of this PR the following tasks were done:

- [x] `LinkButton` component was used in Dropdown Menu instead of `Avatar`.
- [x] `Avatar`'s `onClick` property was marked as deprecated.
- [x] Removed empty `Avatar`s wrapper in `uui/components/widgets` and moved remained tests to `uui-components/src/widgets/__tests__/Avatar.test.tsx`.
- [x] [`Button`, `ButtonBase`] `getProps()` method was added to use corresponding aria attributes. For now such attributes are added when  Button is a dropdown.
- [x] Test snapshots are updated. 

**Before:**
image 
------
**After:**
image

-------
Video with Screen Reader **on**:

https://github.com/epam/UUI/assets/11524637/b939f0d1-bdb2-41b1-8711-54a095bd73df

Note: visible focus during keyboard navigation will be done in a separate PR.


app/src/docs/_examples/dropdownMenu/Basic.example.tsx
```diff
@@ -107,7 +108,15 @@ export default function BasicDropdownMenuExample() {
...
                 closeDelay={ 400 }
                 renderBody={ (props) => renderDropdownBody(props) }
-                renderTarget={ (props) => <Avatar img="https://avatars.dicebear.com/api/human/avatar12.svg?background=%23EBEDF5&radius=50" size="36" { ...props } /> }
+                renderTarget={ (props) => (
+                    <LinkButton
+                        icon={ PersonIcon }
+                        caption="Alex Smith"
+                        size="36"
+                        { ...props }
+                        rawProps={ { 'aria-haspopup': props.isDropdown, 'aria-expanded': props.isOpen } }
```
maryia-lapata, line 117: Here I added a couple of `aria` labels for proper screen reader announcing. Probably it will be better if `Dropdown` component can pass them via `props.rawProps` for example .

AlekseyManetov, line 117: I guess we can add this attributes inside to the Button component, that they will be set by default if component act as dropdown.

AlekseyManetov: The current implementation of `ButtonBase` inheritance makes it impossible to achieve this (we will need to rework it in the future). As a solution, I suggest adding a `getProps` method to the `ButtonBase` that allows you to pass accessibility attributes from the `Button` component.

uui/components/widgets/Avatar.tsx
```diff
@@ -1,4 +1,15 @@
-import { withMods } from '@epam/uui-core';
+import { devLogger, withMods } from '@epam/uui-core';
 import { Avatar as uuiAvatar, AvatarProps } from '@epam/uui-components';
 
-export const Avatar = withMods<AvatarProps>(uuiAvatar, () => []);
+export const Avatar = withMods<AvatarProps>(
+    uuiAvatar,
+    () => [],
+    (props) => {
+        if (__DEV__) {
```
AlekseyManetov, line 8: I propose moving this warning to the Avatar component in `@epam/uui-components'` package and make just re-export here, except one more wrapping with `withMods`

AlekseyManetov (changes requested):

changelog.md
```diff
@@ -9,6 +9,7 @@
 * [DatePicker]: Size '48' marked as @deprecated. It will be removed in future versions. In dev-mode you can see an error if you use this size.
 * Added adapter for react-router 6.
     - Note: we strongly discourage the use of react-router 6, as it introduces too many breaking changes, and certain important features (like block and listen) are available only via unstable internal API.
+* [Avatar]: Property `onClick` marked as @deprecated. It will be removed in future versions. In dev-mode you can see an error if you use this prop.
```
AlekseyManetov, line 12: This version ware already released, please move this mentrion into new draft block(update your brunch with develop)

AlekseyManetov (changes requested):

uui-components/src/buttons/ButtonBase.tsx
```diff
@@ -60,6 +60,8 @@ export abstract class ButtonBase<ButtonProps extends ButtonBaseProps> extends Re
         return this.props.tabIndex || 0;
     }
 
+    getAriaAttrs?(): Record<`aria-${string}`, string | boolean>;
```
AlekseyManetov, line 63: Probably we can make this functions not specific for Aria attributes, that we can pass any Button attributes if needed, it will be more versatile.

AlekseyManetov (changes requested):

Issue #1464, opened 2023-07-05
[Control Group]: improvement according to the design specs


OlegKryvokobylskyi:
### Now
No examples described  with prefix, and suffix with inputs:
![image](https://github.com/epam/UUI/assets/129292145/ebbd5dca-dfde-4467-8d63-7da4ecf90e36)

### To Do
Add examples with prefix, suffix with inputs (then confirm result with design)
https://www.figma.com/file/ozuZOZNqKMUGiwLKYlsfkT/UUI-%E2%80%93-Forms%3A-Text-Input-(Sandbox)?type=design&node-id=616-344660&mode=design&t=L7KKVVAXafMs7NYa-0

NatalliaAlieva:
Released in 5.2.0 version.

PR #1547, opened 2023-07-31
prefix_demo_sanbox_page

Nosafi:
Issue link (if exists):
https://github.com/epam/UUI/issues/1464

Description:
Added sandbox page with prefixes test component.

### DIFF
--- a/app/src/sandbox/prefixes/PrefixComponent/PrefixComponent.tsx
+++ b/app/src/sandbox/prefixes/PrefixComponent/PrefixComponent.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+import './PrefixComponent.scss';
+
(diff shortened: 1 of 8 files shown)


app/src/sandbox/prefixes/PrefixComponent/PrefixComponent.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import './PrefixComponent.scss';
```
AlekseyManetov, line 2: use module imports to import styles into component:
import css from './PrefixComponent.scss';
and then in jsx: className={ css.prefixWrapper }

AlekseyManetov (changes requested):

PR #1549, opened 2023-07-31
Feature/UuiTimePicker

vik753:
### Description: 

Both skins are using the UUI implementation, and loveship design has been changed into the promo implementation, and 48 size is deprecated and will be deleted in the future release.

**Testing notes:** needs to test TimePicker in both skins.


vik753:
Testing notes: needs to test TimePicker in both skins.

DzmitryTrubchyk:
Design Review:
- Need to fix hover state for arrow buttons. I suppose Icon button should be there (or reuse hover, press colors for states from icon button)
image

Issue #699, opened 2022-09-17
[RangeDatePicker] Improve user control over presets


vasilii-kovalev:
## Description

Improve user control over presets, which includes:
* Events listening (click, for example)
* Rendering

## Business case

It is necessary to:
1. Save preset to Local Storage for further usage, only when the appropriate element in the presets list is clicked. If the dates were picked manually and match any preset, the preset shouldn't be updated in Local Storage
2. Mark a preset as selected (see the picture below)
![Screenshot of date range picker body with a list of presets, where the last one is marked as selected - has a check mark left to it and has bolder font weight](https://user-images.githubusercontent.com/10310491/190838123-90e06217-f289-47ca-9444-36c98fd92379.jpg)

### Simple

The simplest solution would include the following steps:
* Add a property `onPresetClick` or `onPresetSelect` to `RangeDatePicker` component with the following signature:
```typescript
// `key` is the key in `presets` object.
onPresetClick?: (key: string, event: React.MouseEvent<HTMLButtonElement>) => void;
```
This solves the first point of the business case.
**Note.** As I can see it, it should be called **after** the dates are updated (like a callback in `this.setState` in React), but you may decide to make this callback the main handler, which would allow/require users to update the dates manually.
* Add a property `selectedPreset` with type like this:
```typescript
// The value is the key from `presets` object.
selectedPreset?: string;
```
This solves the second point of the business case.
When `selectedPreset` matches one of the keys in `presets` object, a unique global CSS class (like `uui-presets-item-selected`) should be added to the appropriate preset element.

### Flexible

A bit more complex, but more flexible from the user's perspective, solution would be to add the ability to render individual presets manually. It can be achieved by:
* Creating and exporting preset element component (like `RangeDatePickerPreset`) with the following type:
```typescript
const RangeDatePickerPreset: React.FC<PresetProps> = () => {};
```
`PresetProps` would include:
1. `onClick` handler (that changes the dates)
2. `isSelected` (computed internally using `selectedPreset` from the previous solution)
    1. (If we don't want to mix the solutions and don't want to introduce `selectedPreset` property) since the users has control over the preset element now, they can add class manually via `cx` property
4. Key from the `presets` object (named as `key` or `presetKey`, for example)
5. Appropriate value of the `presets` object by the key (optionally)
* Providing the date picker component with a property `renderPreset` with the following signature:
```typescript
renderPreset?: (props: PresetProps) => React.Node;
```
In this case, the code solving both points of the business case would look like this:
[...]

PR #1564, opened 2023-08-04
preset in reange date picker from string to rectNode

Nosafi:
Issue link(if exists):
https://github.com/epam/UUI/issues/699

Description:
In RangeDatePicker preset's name from string switched to reactNode.

### DIFF
--- a/app/src/sandbox/rangeDatePicker/datePickerDemo.tsx
+++ b/app/src/sandbox/rangeDatePicker/datePickerDemo.tsx
@@ -1,0 +1,57 @@
+import React, { useState } from 'react';
+import { RangeDatePicker, Text } from '@epam/promo';
+import { rangeDatePickerPresets, RangeDatePickerValue } from '@epam/uui';
+import dayjs from 'dayjs';
+import css from './datePickerDemo.module.scss';
+
+const customPrefixes = {
+    last3Days: {
+        name: <div style={ { backgroundColor: 'blue' } } onClick={ () => console.log('Clicked!') }>Custom prefix</div>,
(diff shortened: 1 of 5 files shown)


uui-components/src/inputs/DatePicker/CalendarPresets.tsx
```diff
@@ -34,8 +34,8 @@ export class CalendarPresets extends React.Component<CalendarPresetsProps> {
         return (
             <div ref={ this.props.forwardedRef } className={ cx(uuiPresets.container, this.props.cx) } { ...this.props.rawProps }>
                 <div className={ uuiPresets.header }>Presets</div>
-                {this.getPresets().map((item) => (
-                    <div key={ item.name } className={ uuiPresets.item } onClick={ () => this.props.onPresetSet(item) }>
+                {this.getPresets().map((item, index) => (
```
AlekseyManetov, line 37: Let's use preset object key as a index, it will be always unique. You can add 'key' for each preset in return value in this.getPresets() function:
```
           presets.push({
                    ...this.props.presets[key].getRange(),
                    name: this.props.presets[key].name,
                    key,
                });
```

app/src/sandbox/rangeDatePicker/datePickerDemo.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { useState } from 'react';
```
AlekseyManetov, line 1: Please remove this sandbox, we don't need it after tests.

AlekseyManetov (changes requested):

Issue #1551, opened 2023-07-31
[Alert]: "onClose" alert does not pop up in Property Explorer when appropriate property selected


OlegKryvokobylskyi:
## Description

"onClose" alert does not pop up in Property Explorer when appropriate property selected
## Steps to Reproduce

1. Open [www.uui.epam.com](https://uui.epam.com/)
2. Open Property Explorer for Alert Component
3. Select "onClose" property 
4. Click on the "X" sign

## Actual result
"onClose" alert does not pop

https://github.com/epam/UUI/assets/129292145/7bc0c4c0-ea9d-4b13-b10a-ec92d64d8a62

## Expected result
"onClose" alert should pop op(the same behavior like in Documentation Tab)

## Environment

* UUI version: 5.1.1
* Browser name: any
* Operating System: any

000000armor:
Hi! I'm interested to help in contributing to this repository and would like to work on this issue. Could you please assign it to me?

AlekseyManetov:
Hi @000000armor, i've assigned it to you. Before start please read our Contribution guide - https://github.com/epam/UUI/blob/main/CONTRIBUTING.md

000000armor:
Btw I noticed that we have same issue in [NotificationCard component](https://github.com/000000armor/UUI/blob/3e8b41e2f70845a4ce20a6270fd780b0377e50c8/app/src/docs/_props/epam-promo/components/overlays/notificationCard.props.tsx#L61). We can edit this issue and fix it in one PR or create new issue. What do you think?

AlekseyManetov:
> @AlekseyManetov Btw I noticed that we have same issue in [NotificationCard component](https://github.com/000000armor/UUI/blob/3e8b41e2f70845a4ce20a6270fd780b0377e50c8/app/src/docs/_props/epam-promo/components/overlays/notificationCard.props.tsx#L61). We can edit this issue and fix it in one PR or create new issue. What do you think?

You can fix it in the same PR, without creating a new issue. Your comment here is enough. Thanks for your attentiveness :)

OlegKryvokobylskyi:
Hi @000000armor , 
Fix verified for both components: 'Alert' and 'NotificationCard' in PE. Thank you. 
But I noticed that this alert does not pop up for 'NotificationCard' in Documentation tab. Could you please fix it?

000000armor:
@OlegKryvokobylskyi opened PR

PR #1567, opened 2023-08-04
[Fix] add alert call in `Alert` and `NotificationCard` component docs

000000armor:
#### Issue link: 
[#1551](https://github.com/epam/UUI/issues/1551)

### Description:
Added `alert` call in with `onClose` callbacks for `NotificationCard` and `Alert` components docs

### DIFF
--- a/app/src/docs/_props/epam-promo/components/overlays/alert.props.tsx
+++ b/app/src/docs/_props/epam-promo/components/overlays/alert.props.tsx
@@ -45,7 +45,7 @@ const SnackbarCardDoc = new DocBuilder<AlertProps>({ name: 'Alert', component: A
     .prop('onClose', {
         examples: [
             {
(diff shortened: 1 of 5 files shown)


changelog.md
```diff
@@ -9,6 +9,8 @@
 
 **What's Fixed**
 
+* [Alert]: Added `alert` call for `onClose`'s prop example
```
AlekseyManetov, line 12: We only include fixes related to our packages and components in the changelog. Documentation and site-related fixes should be excluded from this list. Let's remove those mentions from here.

AlekseyManetov (changes requested):

changelog.md
000000armor: done

AlekseyManetov:
@000000armor Thanks for your conribution!

PR #1591, opened 2023-08-16
[DataTable]: sticky rows

Kuznietsov:
## In progress...


uui/components/tables/DataRowsContainer/DataRows.tsx
```diff
@@ -0,0 +1,68 @@
...
+            </div>
+            {childrenRows.length > 0 && (
+                <div>
+                    <DataRows rows={ childrenRows } renderRow={ renderRow } top={ childrenPinnedTop } />
+                </div>
+            )}
+        </div>
+    );
+}
+export function DataRows<TItem, TId>({
```
jakobz, line 40: We need another name for this component. With just 'DataRows', I'm getting confused with DataRowsProps interface - it clashes with DataRowProps (with singular 'Row') - which have a different meaning.

uui/components/tables/DataRowsContainer/DataRows.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+function DataRowGroup<TItem, TId>({
+    row,
+    childrenRows,
+    renderRow,
+    top = 1,
+}: GroupProps<TItem, TId>) {
+    const rowRef = useRef<HTMLDivElement>();
+    const childrenPinnedTop = row.isPinned ? (top + (rowRef.current?.clientHeight ?? 0)) : top;
```
jakobz, line 26: As we'll get ref.current only after rendering, we won't get correct values for 'top' on the first rendering. However, this works, as we have few re-renders before some rows get stuck. Correct?

Kuznietsov: Yes, that's correct.

uui/components/tables/DataRowsContainer/DataRows.tsx
Kuznietsov: Done.

Kuznietsov:
Merged into feature/sticky-rows-with-scroll-to manually.

PR #1592, opened 2023-08-16
[SlateEditor] Add html serialization

dimaanj:
## Summary

Implements editor value HTML serialization. 

Added 'HTML format' example to the RTE doc


app/src/docs/_examples/richTextEditor/Serialization.example.tsx
```diff
@@ -0,0 +1,102 @@
...
+    videoPlugin(),
+    iframePlugin(),
+    codeBlockPlugin(),
+];
+
+const serializeHTML = createSerializer();
+const deserializeHTML = createDeserializer();
+
+export default function SlateEditorBasicExample() {
+    const [value, setValue] = useState<EditorValue>(migrateSchema(demoData.slateSerializationInitialData));
```
AlekseyManetov, line 43: Should we migrate state here, because we already have such migration inside `SlateEditor`? 
Anyway it will be better if we will replace our demo state with new state variant.

app/src/docs/_examples/richTextEditor/Serialization.example.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    const onEdit = () => {
+        setValue(deserializeHTML(serializedHtml));
+        setType('edit');
+    };
+
+    return (
+        <Panel cx={ css.root }>
+            <FlexRow spacing="18" vPadding="12">
+                {type === 'edit' && (
```
AlekseyManetov, line 65: Let's better use Switch component, like we have at another RTE example
image

AlekseyManetov (changes requested):

app/src/docs/_examples/richTextEditor/Serialization.example.tsx
dimaanj: Changed

app/src/docs/_examples/richTextEditor/Serialization.example.tsx
dimaanj: Replaced

uui-editor/src/plugins/iframePlugin/IframeBlock.tsx
```diff
@@ -29,3 +30,37 @@ export function IframeBlock(props: any) {
...
+
+export function SerializationIframe(props: any) {
+    const { element } = props;
+    const src = element.data.src || element.src;
+    const isPdf = element.data.extension === 'pdf';
+
+    return (
+        <div>
+            <div
+                style={ {
```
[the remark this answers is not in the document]
dimaanj: Discussed offline

NatalliaAlieva:
Released in 5.2.0 version.

PR #1603, opened 2023-08-17
[SlateEditor] Fix zIndex for toolbars

dimaanj:
## Summary 

Removes fixed zIndex for toolbars.


uui-core/src/hooks/useLayer.ts
```diff
@@ -0,0 +1,13 @@
+import { useEffect } from 'react';
+import { useUuiContext } from '../services';
+
+export const useLayer = () => {
+    const context = useUuiContext();
+    const layer = context.uuiLayout?.getLayer();
```
AlekseyManetov, line 6: you will get new layer on each hook call, it's better to use useRef hook to avoid that.

AlekseyManetov (changes requested):

uui-core/src/hooks/useLayer.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const useLayer = () => {
+    const context = useUuiContext();
+    const layer = useRef(context.uuiLayout?.getLayer()).current;
+
+    useEffect(() => {
+        return () => layer && context.uuiLayout?.releaseLayer(layer);
+    }, [context]);
+
+    return layer?.zIndex;
```
AlekseyManetov, line 12: let's return the whole layer object, it will be more versatile and better match with hook name.

AlekseyManetov (changes requested):

PR #1616, opened 2023-08-23
[VirtualList]: Sticky with scrollTo

Kuznietsov:
### In progress...


Kuznietsov:
Merged into feature/pickers-rework manually.

PR #1620, opened 2023-08-28
[LazyDataSource]: skeleton should not appear on filter/sort/search change

Kuznietsov:
### Summary

Skeleton is not appearing any more on filters change.

### DIFF
--- a/uui/components/pickers/DataPickerBody.tsx
+++ b/uui/components/pickers/DataPickerBody.tsx
@@ -2,9 +2,9 @@ import React from 'react';
 import {
     Lens, DataSourceState, isMobile, cx,
 } from '@epam/uui-core';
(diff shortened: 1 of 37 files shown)


uui/components/pickers/DataPickerBody.tsx
```diff
@@ -15,6 +15,7 @@ export interface DataPickerBodyProps extends PickerBodyBaseProps {
     editMode?: 'dropdown' | 'modal';
     searchSize?: ControlSize;
     selectionMode?: 'single' | 'multi';
+    addBlocker?: boolean;
```
AlekseyManetov, line 18: I guess we can remove this prop and add blocker for all cases by default.

AlekseyManetov (changes requested):

NatalliaAlieva:
Released in 5.2.0 version.

PR #1636, opened 2023-09-05
[VirtualList]: scroll to index on fold only when required

Kuznietsov:
### Summary
- Added align = 'top' | 'nearest' for scrollTo functionality.
- Added scrolling to index only when folding element, which is not present in a visible list of rows, but pinned.
- Added fix of adding sticky parent, if scrolled too far. Optimized a solution.
- Added fix of checking elements in explicit mode for records with complex ids.
- Added fix for VirtualList while scrolling out of possible range.
- Added fix of NumericInput, if value = undefined.
- Added example of ScrollTo to the VirtualList docs with `nearest` and `top` align positions for ArrayListView.
- Added example of ScrollTo to the DataSource State docs with `nearest` and `top` align positions for LazyListView.


uui/components/tables/DataRowsContainer/DataRowsGroups.tsx
```diff
@@ -23,10 +23,10 @@ function DataRowsGroup<TItem, TId>({
     top = 1,
 }: DataRowsGroupProps<TItem, TId>) {
     const rowRef = useRef<HTMLDivElement>();
-    const childrenPinnedTop = row.isPinned ? (top + (rowRef.current?.clientHeight ?? 0) - 0.5) : top;
+    const childrenPinnedTop = row.isPinned ? (top + (rowRef.current?.clientHeight ?? 0) - 1) : top;
```
AlekseyManetov, line 26: Let's add comment why we subtract 1 here

AlekseyManetov (changes requested):

app/src/docs/_examples/numericInput/Basic.example.tsx
```diff
@@ -3,12 +3,12 @@ import { FlexCell, LabeledInput, NumericInput } from '@epam/promo';
...
 
 export default function BasicExample() {
-    const [value, onValueChange] = useState(null);
+    const [value, onValueChange] = useState<number>();
 
     return (
         <FlexCell width="auto" cx={ css.container }>
             <LabeledInput label="Default settings">
-                <NumericInput value={ value } onValueChange={ onValueChange } />
+                <NumericInput value={ value } onValueChange={ onValueChange } min={ -15 } />
```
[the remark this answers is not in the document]
Kuznietsov: Done.

uui/components/tables/DataRowsContainer/DataRowsGroups.tsx
Kuznietsov: Done,.

Issue #1652, opened 2023-09-13
[useForm]: form does not recover from isInProgress state when onSave is rejected


cpof-tea:
## Prerequisites

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

## Description

`onSave` sets `isInProgress=true`, but does not set `isInProgress=false` back if `onSave` is rejected

## Steps to Reproduce

[Codesandbox](https://codesandbox.io/s/form-not-recovered-from-in-progress-state-3jprfv?file=/Example.tsx)

1. Type required fields
2. Click Save button
3. Verify isInProgress message

## Actual result

Keeps `isInProgress=true` if `onSave` is rejected

## Expected result

Set `isInProgress=false` if `onSave` is rejected

## Environment

* UUI version: 5.1.2
* Browser name: FF 116.0.3
* Operating System: Ubuntu 20.04

NatalliaAlieva:
Released in 5.2.0 version.

PR #1653, opened 2023-09-13
Reset `isInProgress` when `onSave` is rejected

cpof-tea:
#### Issue link(if exists): 
https://github.com/epam/UUI/issues/1652

### Description:

* Added `handleError`: updates state with `isInProgress=false` and calls `onError` as before
* Added test cases
* Updated changelog


uui-core/src/data/forms/useForm.ts
```diff
@@ -220,6 +220,17 @@ export function useForm<T>(props: UseFormProps<T>): IFormApi<T> {
         return savePromise;
     }, []);
 
+    const handleError = (err?: any) => {
+        const newState: FormState<any> = {
```
AlekseyManetov, line 224: Let's better move new state calculation inside updateFormState callback:
`updateFormState((currentValue) =>({...currentValue, isInProgress: false}))`

AlekseyManetov (changes requested):

cpof-tea:
Closes #1652

PR #1672, opened 2023-09-29
Feature/ts docs gen

siarheiyelin:


app/src/common/docs/BaseDocsBlock.tsx
```diff
@@ -1,136 +1,90 @@
...
-                />
-            </>
-        );
+        const docsGenType = this.getDocsGenType();
+        let content = null;
+        if (docsGenType) {
+            content = (
+                <ApiReferenceItemTableForTypeRef showCode={ true } typeRef={ docsGenType } />
+            );
+        } else if (this.getPropsDocPath()) {
```
AlekseyManetov, line 64: I guess we already replace this.getPropsDocPath() in all places and cade remove this condition branch? This component is used only by us, so we shouldn't worry about other cases.

app/src/common/sidebar/SidebarButton.module.scss
```diff
@@ -15,17 +19,24 @@
...
+        overflow-x: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
     }
 
     &.indent-12 {
         :global(.uui-caption) {
-            margin-left: 12px;
+            margin-left: #{$index-1x}px;
+            max-width: #{$caption-width - $index-1x}px;
```
AlekseyManetov, line 31: If we set a max-width not at each button, but on the whole sidebar container, it will not work?

app/src/docs/Rating.doc.tsx
```diff
@@ -1,10 +1,14 @@
...
 import {
-    BaseDocsBlock, DocExample, EditableDocContent, UUI3, UUI4,
+    BaseDocsBlock, DocExample, EditableDocContent, TDocsGenType, UUI3, UUI4,
 } from '../common';
 
 export class RatingDoc extends BaseDocsBlock {
     title = 'Rating';
+
+    // TODO: no such component in "@epam/uui"
+    override getDocsGenType = (): TDocsGenType => ('@epam/uui-components:RatingProps');
```
AlekseyManetov, line 10: get types from loveship package here. Because it's a skin specific component which will not be moved to the uui package.

app/src/docs/Slider.doc.tsx
```diff
@@ -1,10 +1,14 @@
...
 import {
-    BaseDocsBlock, DocExample, EditableDocContent, UUI3,
+    BaseDocsBlock, DocExample, EditableDocContent, TDocsGenType, UUI3,
 } from '../common';
 
 export class SliderDoc extends BaseDocsBlock {
     title = 'Slider';
+
+    // TODO: no such component in "@epam/uui"
+    override getDocsGenType = (): TDocsGenType => ('@epam/uui-components:SliderBaseProps');
```
AlekseyManetov, line 10: get types from loveship package here. Because it's a skin specific component which will not be moved to the uui package.

app/src/docs/SliderRating.doc.tsx
```diff
@@ -1,10 +1,14 @@
...
 import {
-    BaseDocsBlock, DocExample, EditableDocContent, UUI3,
+    BaseDocsBlock, DocExample, EditableDocContent, TDocsGenType, UUI3,
 } from '../common';
 
 export class SliderRatingDoc extends BaseDocsBlock {
     title = 'SliderRating';
+
+    // TODO: no such component in "@epam/uui"
+    override getDocsGenType = (): TDocsGenType => ('@epam/loveship:SliderRatingProps');
```
AlekseyManetov, line 10: get types from loveship package here. Because it's a skin specific component which will not be moved to the uui package.

app/src/data/index.ts
```diff
@@ -1 +1,15 @@
+import { svc } from '../services';
+
 export * from './apiDefinition';
+export type TAppContext = Awaited<ReturnType<typeof loadAppContext>>;
+export async function loadAppContext() {
```
AlekseyManetov, line 5: Why do we load it via appContext, neither directly in the documents page? It will perform unnecessary api calls for another pages, for example when user open just demo page, he don't need to load api references data.

AlekseyManetov (changes requested):

app/src/common/docs/BaseDocsBlock.tsx
siarheiyelin: done

app/src/common/sidebar/SidebarButton.module.scss
siarheiyelin: done

app/src/docs/Rating.doc.tsx
siarheiyelin: done

app/src/docs/Slider.doc.tsx
siarheiyelin: done

app/src/data/index.ts
siarheiyelin: you are right, done

app/src/docs/SliderRating.doc.tsx
siarheiyelin: done

PR #1673, opened 2023-10-01
Bugfix/1663 dropdown menu button icon bug

dalalvarun:
### Description:
Upon investigation, it was found that isDisabled prop was not being passed to DropdownMenuButton's child IconButton. As a result, even if DropddownMenuButton was disabled, the icon would still look like enabled.


AlekseyManetov:
Hi @dalalvarun, thanks for your investigations and fixes. 
Could you please create your fork branch from develop branch and make a PR to the develop. It will also affect your changelog changes, because we already have a stud for the new version in develop, it better to add your mentions to it.

PR #1697, opened 2023-10-20
Write caption after insert image #1683

### LINKED ISSUE #1683: [RTE]: impossible to sign pictures right after insertion
## Description
Signature "Write a caption" doesn't appear right after inserting picture.

## Steps to Reproduce
1. Open [Rich Text Editor](https://uui.epam.com/documents?id=richTextEditor&mode=doc&skin=UUI4_promo&category=components) component.
2. Click on "insert picture" icon on the sticky tool bar.
3. Insert picture.
4. Try to make a capture under it.

## Actual result
"Write a caption" doesn't appear right after picture insertion.

https://github.com/epam/UUI/assets/57751283/9fc4d420-a6a2-40b0-a1dd-1dc0236c5ec0

## Expected result
Signature "Write a caption" appears only after resizing the picture.

## Environment
* UUI version: 5.1.3
* Browser name: any
* Operating System: any

shubhamt619:
#### Issue link:

https://github.com/epam/UUI/issues/1683 

### Description:
Fixed #1683 by allowing showing caption after image was inserted.

### DIFF
--- a/uui-editor/src/plugins/imagePlugin/ImageElement.tsx
+++ b/uui-editor/src/plugins/imagePlugin/ImageElement.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useRef } from 'react';
+import React, { useEffect, useMemo, useRef, useState } from 'react';
 import {
     Box,
     PlateElement,
@@ -47,27 +47,26 @@ export function ImageElement({
         focused && selected && css.resizeHandleVisible, // for mobile
     ];
 
-    const imageParentRef = useRef(null);
+    const imageRef = useRef(null);
+    const [imageWidth, setImageWidth] = useState(0);
 
     useMediaState();
 
     const [currentWidth] = useResizableStore().use.width();
+
     const isCaptionEnabled = useMemo(() => {
-        let captionEnabled = false;
-        if (currentWidth && typeof currentWidth === 'number') {
-            captionEnabled = currentWidth >= MIN_CAPTION_WIDTH;
-        } else if (currentWidth && (typeof currentWidth === 'string')) {
-            if (imageParentRef.current) {
-                const width = imageParentRef.current.getBoundingClientRect().width;
-                captionEnabled = width >= MIN_CAPTION_WIDTH;
-            }
-        }
+        let width = typeof currentWidth == 'string' ? imageRef?.current.width : currentWidth;
+        let captionEnabled = width >= MIN_CAPTION_WIDTH;
         return captionEnabled;
-    }, [currentWidth]);
+    }, [currentWidth, imageWidth]);
+
+    const imageUpdated = () => {
+        setImageWidth(imageRef.current.width);
+    }
 
     return (
         <PlateElement className={ cx(className) } { ...props }>
-            <figure className={ cx(css.group) } contentEditable={ false } ref={imageParentRef}>
+            <figure className={ cx(css.group) } contentEditable={ false }>
                 <Resizable
                     className={ cx(...aligns) }
                     options={ {
@@ -98,6 +97,8 @@ export function ImageElement({
                                 nodeProps?.className,
                             )
                         }
+                        ref={imageRef}
+                        onLoad={imageUpdated}
                     />
                 </Resizable>
 
--- a/uui-editor/src/plugins/imagePlugin/ImageElement.tsx
+++ b/uui-editor/src/plugins/imagePlugin/ImageElement.tsx
@@ -54,6 +54,8 @@ export function ImageElement({
         let captionEnabled = false;
         if (currentWidth && typeof currentWidth === 'number') {
             captionEnabled = currentWidth >= MIN_CAPTION_WIDTH;
+        } else if (currentWidth && (typeof currentWidth === 'string')) {
+            captionEnabled = true;
         }
 
         return captionEnabled;
--- a/uui-editor/src/plugins/imagePlugin/ImageElement.tsx
+++ b/uui-editor/src/plugins/imagePlugin/ImageElement.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useRef } from 'react';
 import {
     Box,
     PlateElement,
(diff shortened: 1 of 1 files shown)


uui-editor/src/plugins/imagePlugin/ImageElement.tsx
```diff
@@ -54,6 +54,8 @@ export function ImageElement({
         let captionEnabled = false;
         if (currentWidth && typeof currentWidth === 'number') {
             captionEnabled = currentWidth >= MIN_CAPTION_WIDTH;
+        } else if (currentWidth && (typeof currentWidth === 'string')) {
+            captionEnabled = true;
```
AlekseyManetov, line 58: We have a requirement that caption should not appear for images that have width lower than MIN_CAPTION_WIDTH. I guess with your fix, it will not work. I supposed that for this logic it's doesn't matter if which typeof of the width property, we everytime can cast in to number. I guess problem here that we didn't set initial image width on loading, could you check it?

AlekseyManetov (changes requested):

uui-editor/src/plugins/imagePlugin/ImageElement.tsx
shubhamt619: Ok let me update this. I will delete this PR and Raise a New One. Thanks

shubhamt619: @AlekseyManetov

AlekseyManetov: @shubhamt619 you can continue in this one, it's not a problem

shubhamt619: @AlekseyManetov  Thank you for being patient, I have updated the code. Could you please Accept the PR if it's upto the standards and our requirements ?

shubhamt619 (commented):
Completed as per request. Please close the PR if everything looks good to you.
@AlekseyManetov

shubhamt619:
@cm-howard Can someone please review and close this PR ?

uui-editor/src/plugins/imagePlugin/ImageElement.tsx
```diff
@@ -47,21 +47,27 @@ export function ImageElement({
...
+    const imageParentRef = useRef(null);
+
     useMediaState();
 
     const [currentWidth] = useResizableStore().use.width();
     const isCaptionEnabled = useMemo(() => {
         let captionEnabled = false;
         if (currentWidth && typeof currentWidth === 'number') {
             captionEnabled = currentWidth >= MIN_CAPTION_WIDTH;
+        } else if (currentWidth && (typeof currentWidth === 'string')) {
```
AlekseyManetov, line 59: Can we have the same logic for both cases, for string and number? I guess that it's doesn't matter which type of `currentWidth`, it's shouldn't figure out in this logic at all. We always need to have valid `currentWidth` value in editor state and always rely on it. As i understand, the problem here that we don't have correct `currentWidth` right after the initialization, i suggest to add logic which will set it.

AlekseyManetov (changes requested):

uui-editor/src/plugins/imagePlugin/ImageElement.tsx
dimaanj: Seems like caption is not shown after insert stil

shubhamt619:
I am looking into this.
Give me some time.

Will apply a more Robust FIX.

uui-editor/src/plugins/imagePlugin/ImageElement.tsx
AlekseyManetov: @shubhamt619 i don't see any code updates from your since last review

shubhamt619: I was going through the react slate documentation to find the solution. I will work on this today and revert as soon as I get the response.

shubhamt619: @dimaanj @AlekseyManetov Please check now

shubhamt619 (commented):
@AlekseyManetov  Please check now

AlekseyManetov (changes requested):

uui-editor/src/plugins/imagePlugin/ImageElement.tsx
```diff
@@ -47,17 +47,22 @@ export function ImageElement({
...
 
+    const isCaptionEnabled = useMemo(() => {
+        let width = typeof currentWidth == 'string' ? imageRef?.current.width : currentWidth;
+        let captionEnabled = width >= MIN_CAPTION_WIDTH;
         return captionEnabled;
-    }, [currentWidth]);
+    }, [currentWidth, imageWidth]);
+
+    const imageUpdated = () => {
+        setImageWidth(imageRef.current.width);
```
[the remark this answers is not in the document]
shubhamt619: @AlekseyManetov  I agree with you completely , but useResizableStore().use.width() gives 'fit-content' initially when the image is uploaded. So what do you suggest in that case ?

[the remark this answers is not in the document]
AlekseyManetov: Can we try to set image width to the `useResizableStore` on image load? Or if it's seems impossible, we can go with your solution, but make width calculation condition to the more strict:
`let width = currentWidth == 'fit-content' ? imageRef?.current.width : currentWidth;`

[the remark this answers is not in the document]
shubhamt619: ok let me try some updates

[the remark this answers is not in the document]
dimaanj: The following should work. 

```
    const imageRef = useRef(null);
    const [currentWidth, setWidth] = useResizableStore().use.width();

    const isCaptionEnabled = useMemo(() => {
        return typeof currentWidth === 'number' && currentWidth >= MIN_CAPTION_WIDTH;
    }, [currentWidth]);

    const imageUpdated = () => {
        setWidth(imageRef.current.width);
    };
```

AlekseyManetov:
@shubhamt619 I'll close this PR, because we finished it in #1744. Thanks for your research and contribution.

Issue #1610, opened 2023-08-21
[Presets]: add the ability to add a modal confirmation window when deleting a preset


NatalliaAlieva:
## Description
Please add a confirmation modal window, since User may accidentally delete a preset without the possibility of restoring it.

AlekseyManetov:
I propose that we modify the `onPresetDelete` function to return a promise. Only after this promise is resolved, we will proceed with the preset deletion on the frontend. This approach allows those who need it to implement a modal confirmation dialog. If the action is canceled within the promise (using reject), no deletion will occur. For those who don't require this extra step, they can simply call the API in this callback directly.

DzmitryTrubchyk:
**Text for modal**

Title: Delete preset
Text: Preset_name will be deleted and can't be restored.

Actions: Cancel/Delete

NatalliaAlieva:
Released in 5.4.0 version.

Issue #1601, opened 2023-08-17
[LabeledInput]: add possibility to customize validationMessage with rich text 


ghost:
## Now
Built-in validation message supposed to be used as string property to a component, so there is not opportunity to customise it, eg. include link to support. 
_Validation message example_: **Something went wrong please contact [support.com](/#)**.

## To Do
Introduce option for the customisation of validation messages.

## Describe the solution you'd like (optional)
Could be an option with the custom rendering or simply changing type to ReactNode.

NatalliaAlieva:
Released in 5.4.0 version.

PR #1706, opened 2023-10-25
[PresetsPanel] Add modal example for presets

dimaanj:
## Summary 

#1610
#1601

- Adds confirmation modal for presets example.
- Change `validationMessageType` for `LabeledInput`

### DIFF
--- a/uui-core/src/hooks/useTableState/useTableState.ts
+++ b/uui-core/src/hooks/useTableState/useTableState.ts
@@ -129,7 +129,7 @@ export const useTableStateImpl = <TFilter = Record<string, any>, TViewState = an
     }, [createPreset, getNewPresetOrder]);
 
     const deletePreset = useCallback(
-        async (preset: ITablePreset<TFilter, TViewState>) => {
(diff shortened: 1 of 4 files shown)


AlekseyManetov (changes requested):

uui-core/src/hooks/useTableState/useTableState.ts
```diff
@@ -128,14 +128,28 @@ export const useTableStateImpl = <TFilter = Record<string, any>, TViewState = an
...
+            const removePreset = () => {
+                params.onValueChange((val) => ({
+                    ...val,
+                    presetId: undefined,
+                }));
+                setPresets((prevValue) =>
+                    prevValue.filter((p) => p.id !== preset.id));
+            };
+
+            if (params.onPresetDelete) {
```
AlekseyManetov, line 142: i guess, we can make it shorter:
```
try {
    await params?.onPresetDelete(preset);
} catch (e) {return;}
removePreset();
```

dimaanj: discussed offline

Issue #1634, opened 2023-09-04
[DropdownMenuButton]: incorrect behavior with isActive for same pathname and different search value


nadiushkaNA:
## Steps to Reproduce
![2023-09-04_12h23_20](https://github.com/epam/UUI/assets/11243577/3da30fb1-1a54-425d-b6f6-45c91d3d060c)

https://codesandbox.io/s/uui-forked-53cvyf?file=/Example.tsx

## Actual result

highlighted all items with same pathname if use link, using isActive props don't help

## Expected result

isActive props should be prioritised

NatalliaAlieva:
Released in 5.4.0 version.

PR #1719, opened 2023-11-01
[DropdownMenuButton] Fix `isActive`

dimaanj:
## Summary

#1634

fix `isActive`


uui-components/src/buttons/ButtonBase.tsx
```diff
@@ -66,7 +66,7 @@ export abstract class ButtonBase<ButtonProps extends ButtonBaseProps> extends Re
         return !!link;
     }
 
-    render(): any {
+    getRenderParams() {
```
AlekseyManetov, line 69: let's rename to 'getAnchorParams'

AlekseyManetov (changes requested):

PR #1723, opened 2023-11-01
[DataTable] Fix move columns in data table when there are hidden columns

dimaanj:
## Summary

Fix columns reorder.

**Before:** when reorder columns, hidden columns start to appear 
**After:** when reorder columns, hidden remain hidden.


NatalliaAlieva:
Released in 5.4.0 version.

PR #1733, opened 2023-11-07
Fix `DataTable` missing `role=table`

### LINKED ISSUE #1732: [DataTable]: missing `role=table` attribute
## Prerequisites

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

## Description

`DataTable` missing `role=table` attribute

## Steps to Reproduce

[Codesandbox](https://codesandbox.io/s/datatable-missing-role-table-ll2h78)

1. Verify console output for `querySelector("[role=table]")` assertion or check with devtools manually

## Actual result

No presence of `role=table`

## Expected result

Presence of `role=table`

## Environment

* UUI version: 5.2.0
* Browser name: FF 119.0
* Operating System: Ubuntu 20.04

## Comment

Issue introduced in `5.2.0`

cpof-tea:
#### Issue link(if exists):

Resolves #1732

### Description:

Fixes `DataTable` missing `role=table`


uui-components/src/layout/ScrollBars.tsx
```diff
@@ -6,7 +6,7 @@ import { IHasCX, cx, IHasRawProps } from '@epam/uui-core';
 import css from './ScrollBars.module.scss';
 import type { Scrollbars, ScrollbarProps as LibScrollbarProps, positionValues } from 'react-custom-scrollbars-2';
 
-export interface ScrollbarProps extends IHasCX, Omit<LibScrollbarProps, 'ref'>, IHasRawProps<Scrollbars> {
+export interface ScrollbarProps extends IHasCX, Omit<LibScrollbarProps, 'ref'>, IHasRawProps<LibScrollbarProps> {
```
cpof-tea, line 9: I think `rawProps` meant to be scrollbar props,

'cause `Scrollbars` is actually a component type, meaning right now `rawProps` accepts every component property and method, for example it will accept render method:
2023-11-07_20-11-15_1699374975_1999x1374_scrot
and at [L74](https://github.com/epam/UUI/pull/1733/files#diff-88877c2e35d928cfefe91b1e85d4d652922d49c90b98c0a290ce65cc9d9dcc3fL74) `rawProps` spreads to scrollbar props.

So, it definitely meant to be scrollbar props

uui/components/layout/VirtualList.tsx
```diff
@@ -30,7 +30,7 @@ type VirtualListRenderRows<List extends HTMLElement = any> = {
 interface BaseVirtualListProps
     extends IHasCX,
     IEditable<VirtualListState>,
-    IHasRawProps<HTMLAttributes<HTMLDivElement>> {
+    IHasRawProps<ScrollbarProps> {
```
cpof-tea, line 33: Since `ScrollbarProps` already has all `HTMLAttributes`, and since `rawProps` passed as is to nested `Scrollbars` component, it's better to set `rawProps` type to `ScrollbarProps` instead, to fix inconsistency between overlapping DOM methods of `HTMLDivElement` and `ScrollbarProps` interfaces

uui-components/src/layout/ScrollBars.tsx
```diff
@@ -6,7 +6,7 @@ import { IHasCX, cx, IHasRawProps } from '@epam/uui-core';
 import css from './ScrollBars.module.scss';
 import type { Scrollbars, ScrollbarProps as LibScrollbarProps, positionValues } from 'react-custom-scrollbars-2';
 
-export interface ScrollbarProps extends IHasCX, Omit<LibScrollbarProps, 'ref'>, IHasRawProps<Scrollbars> {
+export interface ScrollbarProps extends IHasCX, Omit<LibScrollbarProps, 'ref'>, IHasRawProps<React.HTMLAttributes<Scrollbars['container']>> {
```
cpof-tea, line 9: I think `rawProps` meant to be scrollbar container props (which is typeof `HTMLDivElement`),

'cause `Scrollbars` is actually a component type, meaning right now `rawProps` accepts every component property and method, for example it will accept render method:
2023-11-07_20-11-15_1699374975_1999x1374_scrot
and at [L74](https://github.com/epam/UUI/pull/1733/files#diff-88877c2e35d928cfefe91b1e85d4d652922d49c90b98c0a290ce65cc9d9dcc3fL74) `rawProps` spreads to scrollbar props.

uui-components/src/layout/ScrollBars.tsx
```diff
@@ -71,7 +71,7 @@ export const ScrollBars = forwardRef<ScrollbarsApi, ScrollbarProps>(({
             hideTracksWhenNotNeeded
             ref={ bars }
             { ...props }
-            { ...rawProps }
+            { ...(rawProps as unknown as LibScrollbarProps) }
```
cpof-tea, line 74: `LibScrollbarProps` wrongly typed in the library itself,
 
`HTMLAttributes<Scrollbars>` incompatible with `HTMLAttributes<HTMLDivElement>` - it has conflicts in DOM methods, since `Scrollbars` is a React component type, and not inherits `HTMLElement`

So, we should locally patch package or transform type this way by force

uui-components/src/layout/ScrollBars.tsx
cpof-tea: PR RobPethick/react-custom-scrollbars-2#49, bot not sure it will be merged and released, since repository contains stale pending PRs. Will ping the maintainer

uui/components/layout/VirtualList.tsx
```diff
@@ -81,6 +81,7 @@ export const VirtualList = React.forwardRef<ScrollbarsApi, VirtualListProps>((pr
             onScroll={ handleScroll }
             renderView={ ({ style }: any) => <VirtualListView isLoading={ props.isLoading } style={ style } /> }
             ref={ scrollBarsRef }
+            rawProps={ props.rawProps }
```
AlekseyManetov, line 84: I guess that we may not to pass rawProps to the `ScrollBars`, due to we have VirtualListView component which is a wrapper for virtual list content. This component already has `rawProps` prop, but it's needed to pass them in 82 line.

AlekseyManetov (changes requested):

uui/components/layout/VirtualList.tsx
cpof-tea: Yeah, I thought about that in the first place, but I didn't like that element with `class=table` is not receiving `role=table`, but its' child does

AlekseyManetov: I think it's not a problem. Moreover, I supposed that the second solution will be more correct, because it that case role=table will be on wrapper which directly contains header and rows

AlekseyManetov:
@cpof-tea thanks for your contribution!

PR #1734, opened 2023-11-08
[DataTable] Fix checkboxes when pin columns

dimaanj:
## Summary

Fix checkboxes when pin columns:
- now `displayedPinned` group only contains columns with`fix: left`
- pin / unpin action will always place column before `displayedUnpinned` list (this change fixes issue).

**Before:** when pin the first column, checkboxes remains in the prev one
**After:**  when pin the first column, checkboxes appear in the left (just pinned) column


NatalliaAlieva:
Released in 5.4.0 version.

PR #1758, opened 2023-11-17
[DataTable]: fixed `Select All Pages` of table with paging.

Kuznietsov:
### Summary

Added possibility to select all records of the table (ignoring pages).


app/src/docs/_examples/tables/PagedTable.example.tsx
```diff
@@ -39,37 +47,46 @@ export default function PagedTable() {
...
+                columns={ columns }
+                headerTextCase="upper"
+            />
             <FlexRow size="36" padding="12">
                 <FlexSpacer />
                 <Paginator
                     value={ state.page }
-                    onValueChange={ (newPage) => setState({ ...state, page: newPage, scrollTo: { index: 0 } }) }
-                    totalPages={ Math.ceil(state.totalCount / state.pageSize) }
+                    onValueChange={ (newPage) => setTableState({ ...state, page: newPage, scrollTo: { index: 0 } }) }
```
jakobz, line 88: Setting scrollTo can now be moved to the new setTableState function, to have all state-setting logic at one place.

app/src/docs/_examples/tables/PagedTableWithSelectAll.example.tsx
```diff
@@ -0,0 +1,100 @@
...
+    }, [api, state]);
+
+    const dataSource = useLazyDataSource<Person, number, unknown>({
+        api,
+        rowOptions: {
+            checkbox: {
+                isVisible: true,
+            },
+        },
+        backgroundReload: true,
```
jakobz, line 65: Isn't backgroundReload enabled by-default?

uui-core/src/data/processing/views/tree/BaseTree.ts
```diff
@@ -148,6 +148,11 @@ export abstract class BaseTree<TItem, TId> implements ITree<TItem, TId> {
     }
 
     public getTotalRecursiveCount() {
+        const rootInfo = this.nodeInfoById.get(undefined);
```
jakobz, line 151: I think we either should have either of:
A) Keep all totalCount-related hacks in getListProps(), and keep Tree.getTotalRecursiveCount straightforward (it should sum all counts of all leafs if all counts are known)
B) Move most logic into the Tree.getTotalRecursiveCount

I vote for A, as we have more data in getListProps, and Tree.getTotalRecursiveCount returning value bigger than we actually have in the Tree structure (in case of paging) - looks very misleading. I'd rather don't use totalCount in the Tree at all, and threat this as some generic opaque list info.

uui-core/src/data/processing/views/tree/LoadableTree.ts
```diff
@@ -249,6 +249,10 @@ export abstract class LoadableTree<TItem, TId> extends EditableTree<TItem, TId>
             if (newNodesCount !== nodeInfo.count) {
                 nodeInfo = { ...nodeInfo, count: newNodesCount };
             }
+
+            if (parent == null && response.totalCount != null && nodeInfo.totalCount !== response.totalCount) {
```
jakobz, line 253: nodeInfo.totalCount !== response.totalCount check looks unnecessary here

jakobz: Probably we don't need this 'if' at all - I don't see why should we ignore totalCount on non-root nodes.

One case I see, is that API might want to query server for totalCount once per list, and don't re-query it for each page/each (topIndex, visibleCount). However, if we allow this, e.g. by assuming totalCount=undefined to mean "keep previous value", I'm afraid of some stale-cache issues...

app/src/docs/_examples/tables/PagedTableWithSelectAll.example.tsx
```diff
@@ -0,0 +1,100 @@
...
+            const result = await svc.api.demo.personsPaged({
+                ...rq,
+                filter: { departmentId: 13 }, // to get less results and non round-numbered number of people
+            });
+            return result;
+        },
+        [svc.api.demo],
+    );
+
+    const selectAll = useCallback(async (shouldSelectAll: boolean) => {
```
jakobz, line 47: Actually, this work-around looks completely legit and straightforward solution to the problem.

jakobz (changes requested):
Suggested some improvements, see comments

app/src/docs/_examples/tables/PagedTableWithSelectAll.example.tsx
Kuznietsov: Unfortunately, no, it is not.

app/src/docs/_examples/tables/PagedTable.example.tsx
Kuznietsov: `setTableState` is used in other places too, so, it can cause overriding of a new state in case of `pageSize` ~ 100 (like, you won't be able to scroll to 21'st item, because scrollTo has a new link and VirtualList will be forced to scroll to that position.

https://github.com/epam/UUI/blob/6fa0ee3a88f345ce8433d730f8e46147d567bc69/app/src/docs/_examples/tables/PagedTable.example.tsx#L80
https://github.com/epam/UUI/blob/6fa0ee3a88f345ce8433d730f8e46147d567bc69/app/src/docs/_examples/tables/PagedTable.example.tsx#L71


@jakobz, WDYT?

uui-core/src/data/processing/views/tree/LoadableTree.ts
Kuznietsov: @jakobz, it makes sense. Done.

app/src/docs/_examples/tables/PagedTable.example.tsx
jakobz: Got it. Let's keep as is.

app/src/docs/_examples/tables/PagedTableWithSelectAll.example.tsx
jakobz: Ok. Probably we'll enable this by default later.

app/src/docs/tables/Advanced.doc.tsx
```diff
@@ -12,13 +12,14 @@ export class AdvancedTablesDoc extends BaseDocsBlock {
 
                 <DocExample cx={ css.appBg } title="Columns Configuration" path="./_examples/tables/ColumnsConfig.example.tsx" />
 
-                <DocExample cx={ css.appBg } title="Table with rows Drag&Drop" path="./_examples/tables/TableWithDnD.example.tsx" />
+                <DocExample title="Table with rows Drag&Drop" path="./_examples/tables/TableWithDnD.example.tsx" />
```
AlekseyManetov, line 15: Add cx={ css.appBg } for all examples here, it's required that table wrapper have background according to the selected theme

AlekseyManetov (changes requested):

PR #1780, opened 2023-11-27
Feature/Tag new colors by design

vik753:


uui/components/types.ts
```diff
@@ -26,6 +26,11 @@ export const allEpamBadgeSemanticColors: EpamBadgeSemanticColor[] = [
     'info', 'success', 'warning', 'critical', 'neutral',
 ];
 
+export type EpamTagSemanticColor = 'info' | 'success' | 'warning' | 'critical' | 'neutral';
```
AlekseyManetov, line 29: let's use `SemanticColor` type and just add 'neutral':
`export type EpamTagSemanticColor = SemanticColor | 'neutral';`
also let's fix it for `EpamBadgeSemanticColor`.

Also, i guess we don't need variables like  `allEpamTagSemanticColors`, since now we generate PE base on prop type. I guess we use them only for the PE configuration

uui/components/widgets/Tag.module.scss
```diff
@@ -1,13 +1,15 @@
...
+    --uui-tag-bg: var(--uui-color-50);
+    --uui-tag-bg-hover: var(--uui-color-60);
+    --uui-tag-border: var(--uui-color-50);
+    --uui-tag-border-hover: var(--uui-color-60);
+    --uui-tag-caption: var(--uui-neutral-0);
+    --uui-tag-fill: var(--uui-neutral-0);
+    --uui-tag-fill-hover: var(--uui-tag-fill);
 
-    --uui-tag-border-radius: 30px;
+    --uui-tag-border-radius: 3px;
```
AlekseyManetov, line 12: use theme specific --uui-border-radius variable as a default value

AlekseyManetov (changes requested):

uui/components/types.ts
vik753: fixed

uui/components/widgets/Tag.module.scss
vik753: fixed

app/src/demo/tables/filteredTable/columns.tsx
```diff
@@ -20,7 +20,7 @@ export const personColumns: DataColumnProps<Person, number>[] = [
         render: (p) =>
             p.profileStatus && (
                 <FlexRow>
-                    <Badge size="24" indicator fill="outline" color={ p.profileStatus.toLowerCase() as BadgeColor } caption={ p.profileStatus } />
+                    <Badge size="24" indicator fill="outline" color={ p.profileStatus.toLowerCase() as BadgeColors } caption={ p.profileStatus } />
```
siarhei-epam, line 23: use singular noun for singular props

vik753: fixed

siarhei-epam (approved):

PR #1815, opened 2023-12-05
[Inputs] Test inputs folder

dimaanj:
## Summary

Inprove tests coverage for inputs


uui-components/src/inputs/__tests__/Checkbox.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    it('should handle focus event', () => {
+        const onFocus = jest.fn();
+        render(<Checkbox value={ false } onValueChange={ jest.fn } onFocus={ onFocus } />);
+        const input = screen.getByRole('checkbox');
+
+        input.focus();
+
+        expect(onFocus).toHaveBeenCalled();
```
AlekseyManetov, line 57: Let's also check that input receive focus, at least that checkbox element have '.uui-focus' class. Or probably we can understand it via some other element props. Probably we can use something like this expect(input).toHaveFocus();. Check this for all other component tests in this PR.

uui-components/src/inputs/__tests__/Checkbox.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+    it('should handle blur event', () => {
+        const onBlur = jest.fn();
+        render(<Checkbox value={ false } onValueChange={ jest.fn } onBlur={ onBlur } />);
+        const input = screen.getByRole('checkbox');
+
+        input.focus();
+        input.blur();
+
+        expect(onBlur).toHaveBeenCalled();
```
AlekseyManetov, line 68: The same as for focus, let's not only check that user callback was called but also ensure that component works properly.

uui-components/src/inputs/__tests__/Checkbox.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+            return <Checkbox { ...props } />;
+        },
+    );
+    return {
+        setProps,
+        mocks,
+    };
+}
+
+describe('Checkbox', () => {
```
AlekseyManetov, line 24: Let's also check that click on label toggle checkbox value and that user can change value via keyboard

uui-components/src/inputs/__tests__/RadioInput.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+            return <RadioInput { ...props } />;
+        },
+    );
+    return {
+        setProps,
+        mocks,
+    };
+}
+
+describe('RadioInput', () => {
```
AlekseyManetov, line 24: Let's also check focus/blur + label click + changing value via keyboard

uui-components/src/inputs/__tests__/Switch.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+            return <Switch { ...props } />;
+        },
+    );
+    return {
+        setProps,
+        mocks,
+    };
+}
+
+describe('Switch', () => {
```
AlekseyManetov, line 24: Let's also check focus/blur + label click + changing value via keyboard

AlekseyManetov (changes requested):

PR #1857, opened 2023-12-17
[DatePicker] add date picker tests

dimaanj:
## Summary

Adjust date picker tests


uui/components/datePickers/RangeDatePickerBody.tsx
```diff
@@ -21,14 +21,14 @@ export function weekCount(displayedDate: Dayjs) {
     return arrayToMatrix(days, 7).length;
 }
 
-const uuiRangeDatePickerBody = {
+export const uuiRangeDatePickerBody = {
```
AlekseyManetov, line 24: we need to avoid exporting this internal variable from the package

uui/components/datePickers/RangeDatePickerBody.tsx
```diff
@@ -88,28 +88,29 @@ export class RangeDatePickerBody extends React.Component<RangeDatePickerBodyProp
...
+        const fromValue = this.props.value.selectedDate?.from ? dayjs(this.props.value.selectedDate.from).valueOf() : null;
+        const toValue = this.props.value.selectedDate?.to ? dayjs(this.props.value.selectedDate.to).valueOf() : null;
 
-        const inRange = dayValue >= fromValue && dayValue <= toValue && fromValue !== toValue && fromValue && toValue;
+        const inRange = fromValue && toValue && dayValue >= fromValue && dayValue <= toValue && fromValue !== toValue;
         const isFirst = dayValue === fromValue;
         const isLast = dayValue === toValue;
 
-        return [
+        return [cx(
```
AlekseyManetov, line 100: Are you sure that wee ned to have `cx` call here? I guess the `cx` will be called under the result of this function in place we `getDayCX` was called

uui/components/datePickers/__tests__/RangeDatePickerBody.test.tsx
```diff
@@ -173,14 +192,80 @@ describe('RangeDatePickerBody', () => {
         const [oct13] = useCase2.result.queryAllByText('13');
         const [oct14] = useCase2.result.queryAllByText('14');
 
-        function parentElemContainsClasses(elem: HTMLElement, classesArr: string[]) {
-            // @ts-ignore
-            const actualList = [...elem.parentElement.classList];
-            return classesArr.every((c: string) => actualList.indexOf(c) !== -1);
-        }
+        expect(parentElemContainsClasses(oct12, [uuiRangeDatePickerBody.firstDayInRangeWrapper, uuiDaySelection.selectedDay])).toBeTruthy();
```
AlekseyManetov, line 195: It's not a good practice to rely on css classes in tests. Probably, we can add for this nodes some html attributes(may be from a11y) and then use them in tests except using classnames?

AlekseyManetov (changes requested):

uui/components/datePickers/__tests__/DatePicker.test.tsx
```diff
@@ -4,24 +4,23 @@ import {
 } from '@epam/uui-test-utils';
 import { DatePicker, DatePickerProps } from '../DatePicker';
 
-async function setupDatePicker(params: Partial<DatePickerProps>) {
+async function setupDatePicker(params: Pick<DatePickerProps, 'value' | 'format' | 'filter'>) {
```
AlekseyManetov, line 7: Let's also add tests for:
 - entering date from keyboard in target input. It will also support entering date not only in default format, but also in some others supported by DatePicker formats.
- isHoliday prop

uui/components/datePickers/__tests__/RangeDatePicker.test.tsx
```diff
@@ -13,21 +13,20 @@ async function setupRangeDatePicker(params: { value: { from: string; to: string
             value: value ?? { from: null, to: null },
```
AlekseyManetov, line 13: Let's also add tests for:
 - entering date from keyboard in target input. It will also support entering date not only in default format, but also in some others supported by DatePicker formats.
 - moving focus in inputs depends on which part are selecting now
 - onFocus/onBlur props
 - isHoliday prop
 - onOpenChange prop

AlekseyManetov (changes requested):

uui/components/datePickers/RangeDatePickerBody.tsx
dimaanj: getDayCX should be passed as `(day: Dayjs): string[];`. but currently we have `(day: Dayjs) => (string | false | 0 | null)[]`. To meet the type I call `cx` here.

uui/components/datePickers/__tests__/RangeDatePickerBody.test.tsx
dimaanj: discussed offline

uui/components/datePickers/__tests__/DatePicker.test.tsx
```diff
@@ -103,4 +109,42 @@ describe('DatePicker', () => {
...
+        const { dom, mocks } = await setupDatePicker({ value: null, format: DATE_FORMAT_DEFAULT });
+        expect(dom.input.value).toEqual('');
+        fireEvent.change(dom.input, { target: { value: '2017-01-22' } });
+        fireEvent.blur(dom.input);
+        expect(mocks.onValueChange).toHaveBeenCalledWith('2017-01-22');
+        expect(dom.input.value).toEqual('Jan 22, 2017');
+    });
+
+    it('should support entering date from keyboard in custom format', async () => {
+        const { dom, mocks } = await setupDatePicker({ value: null, format: DATE_FORMAT_CUSTOM });
```
AlekseyManetov, line 123: format: DATE_FORMAT_CUSTOM doesn't affect in which format date should be entered in input. We have a support of some of popular formats which hardcoded in DatePicker. Therefore this tests should work without format prop, let's remove it. And let's also test a few more supported formats, you can merge this test with previous one, because that check the similar behavior.

This comment also related for RangeDatePicker tests

uui/components/datePickers/__tests__/DatePicker.test.tsx
```diff
@@ -103,4 +109,42 @@ describe('DatePicker', () => {
         expect(mocks.onValueChange).toHaveBeenCalledWith('2017-01-31');
         expect(dom.input.value).toEqual('31-01-2017');
     });
+
+    it('should support entering date from keyboard in default format', async () => {
+        const { dom, mocks } = await setupDatePicker({ value: null, format: DATE_FORMAT_DEFAULT });
+        expect(dom.input.value).toEqual('');
+        fireEvent.change(dom.input, { target: { value: '2017-01-22' } });
```
AlekseyManetov, line 116: You are entering date not in default format.

uui/components/datePickers/__tests__/RangeDatePicker.test.tsx
AlekseyManetov: Still doesn't find test for `onOpenChange` prop and 'moving focus in inputs depends on which part are selecting now' depends on selection in picker body, not only in cases when some input was clicked directly.

AlekseyManetov (changes requested):

uui/components/datePickers/__tests__/RangeDatePicker.test.tsx
dimaanj: Added tests for `focus`change.

uui/components/datePickers/__tests__/DatePicker.test.tsx
dimaanj: Yes, I intended to check that after outside click format will be changed to one we provided.

uui/components/datePickers/__tests__/DatePicker.test.tsx
dimaanj: Added tests with formats.

PR #1880, opened 2023-12-22
[useTree]: hook

Kuznietsov:
### New features

| Feature/class/hook/field                     | Description                                                                                                                                          |
| -------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
| __BaseListViewProps.showSelectedOnly__       | Enables returning of only selected rows from __IDataSource.useView__                                                                                 |
| __useTree__                                  | Hook, which is used to create trees of different types with further <br/>access to loaded data                                                       |
| __useLazyFetchingAdvisor__                   | Hook, which is used to advice if data should be reloaded/refetched/<br/>some rows should be uploaded additionally                                    |
| __ItemsStorage__                             | A storage of items, loaded by trees. Is used to share data between <br/>multiple trees/views                                                         |
| __useItemsStorage__                          | Hook for creating new ItemsStorage or using the existing itemsMap/setItems                                                                           |
| __ItemsMap__                                 | Immutable map for storing items.                                                                                                                     |
| __ITree__                                    | Tree interface, which is used in useDataRows. It can be<br/> implemented by users, if they don't want to use our trees,<br/> but want to render rows |
| __useDataRows__                              | Hook, which builds rows                                                                                                                              |
| __useCascadeSelectionService__               | Hook, which allows to perform cascadeSelection for lazy trees.                                                                                       |
| __BaseListViewProps.patchItems__             | Map of items to be added to the provided dataset in a tree.                                                                                          |
| __BaseListViewProps.isDeleted__              | Getter for detecting if an item is deleted.                                                                                                          |
| __BaseListViewProps.fixItemBetweenSortings__ | If enabled, positions of items between sorting changes are fixed.                                                                                    | 
| __BaseListViewProps.getNewItemPosition__     | Function, which returns an relative position of a new item.                                                                                          |
| __BaseListViewProps.getItemTemporaryOrder__  | Function, which returns a temporary order of a new item.                                                                                             |
| __BaseListViewProps.sortBy__                 | Function, which provides a value of a field to be sorted by.                                                                                         |

### API Updates

- __IDataSource.useView__ accepts new __onValueChange__ function, which should handle updateValue callback.
- __DataTableProps__ accepts additional  __rows__ prop, along with __getRows()__.
- __ITableState.setTableState__ function should handle both, value and updateValue callback.

### Deprecation

| Deprecated                                   | Suggestions                                                                                                                                      |
| -------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------ |
| __ArrayListView__                            | Use __ArrayDataSource.useView__ instead                                                                                                          |
| __AsyncListView__                            | Use __AsyncDataSource.useView__ instead                                                                                                          |
| __LazyListView__                             | Use __LazyDataSource.useView__ instead                                                                                                           |
| __useList__                                  | Use __IDataSource.useView__ instead                                                                                                              |
| __IDataSource.unsubscribeView__              | This functionality is built in __IDataSourceView.useView__ hook                                                                                  |
| __IDataSource.getView__                      | Use __IDataSource.useView__ instead                                                                                                              |
| __IDataSourceView.activate__                 | This functionality is built in __IDataSourceView.useView__ hook                                                                                  |
| __IDataSourceView.deactivate__               | This functionality is built in __IDataSourceView.useView__ hook                                                                                  |
| __IDataSourceView.getSelectedRows__          | Pass __showSelectedOnly = true__ to __IDataSourceView.useView__ <br/> options and access selected rows via __IDataSourceView.getVisibleRows()__. | 
| __IDataSourceView.loadData__                 | -                                                                                                                                                |
| __IDataSourceView.\_forceUpdate__            | -                                                                                                                                                |
| __LazyListViewProps.legacyLoadDataBehavior__ | -                                                                                                                                                |


app/src/docs/_examples/tables/ArrayTable.example.tsx
```diff
@@ -1,20 +1,23 @@
...
-    );
+    const { tree, ...dataRowProps } = useTree({
+        getId: (item) => item.id,
+        dataSourceState,
+        setDataSourceState,
+        items: demoData.featureClasses,
+    }, []);
 
-    const view = dataSource.useView(value, onValueChange, {});
+    const { listProps, visibleRows } = useDataRows({
```
jakobz, line 17: Let's try to adjust useDataRows output and DataTable's props to be compatible, and more convenient.

My proposal:
```
const { rowsCount, rows, totalCount } = useDataRows(); // to use separately

// to pass-thru to the table: 
const rowProps = useDataRows();
<DataTable { ...rowsProps } />
```
DataTable needs a minimal adjustments to allow ```rows``` prop, as alternative to getVisibleRows.

I also believe that ```rows``` is better than ```visibleRows```, as 'visible' doesn't add any clarification in this case.

uui-core/src/data/lenses/lensesImpl.tsx
```diff
@@ -58,6 +59,33 @@ export function prop<TObject, TKey extends keyof TObject>(name: TKey): ILensImpl
     };
 }
 
+export function get<TItem, TId>(id: TId): ILensImpl<ItemsMap<TId, TItem>, TItem> {
```
jakobz, line 62: Let's:
- generalize this Lens to support any Map-like object. We can re-use the trick we used to abstract Map in Tree via IMap interface. This approach allows to support a native Map, as well as other Map-like objects. Including immutable Maps like I.Map - they can implement constructor as no-op.
- we need a better name for both ILensImpl and LensBuilder method. I suggest ```mapItem```, but let's discuss - probably there'll be a better ideas.

uui-core/src/data/processing/ItemsMap.ts
```diff
@@ -0,0 +1,96 @@
+export type OnUpdate<TId, TItem> = (newItemsMap: ItemsMap<TId, TItem>) => void;
+
+interface ModificationOptions {
```
jakobz, line 3: We can be more explicit in this API - just tell callee how we got these items:

```typescript
interface ModificationOptions {
  /** Where new/updated items came from:
  * api - received them from the recent api call
  * edit - received from dataRow.onValueChange callback
  */
  source: 'api' | 'edit'; 
}
```

Users can then decide what to do in each case:

```typescript
setItems((updatedItems, options) => {
   let items = new Map(myState.items);   
   if (options.source === 'api') {
      items = new Map([...updatedItems, ...items]) // items from state take precedence, to not override edited items
   } else {
      items = new Map([...items, ...updatedItem]) // items from state take precedence, to not override edited items
   }
})

File: app/src/docs/_examples/tables/ArrayTable.example.tsx (hunk 1 above)
Comment (Kuznietsov) [reply app/src/docs/_examples/tables/ArrayTable.example.tsx:17]: I agree on `rows`. It can be aligned without any problem.
There is a problem with the `pass-through` approach.
```ts
interface DataRowsResult<TItem, TId> {
  rows: DataRowProps<TItem, TId>[];
  getSelectedRows: : ({ topIndex, visibleCount }?: VirtualListRange) => DataRowProps<TItem, TId>[];
  getSelectedRowsCount: : () => number;
  getById:  (id: TId, index: number) => DataRowProps<TItem, TId>;
  clearAllChecked: () => void;
        
  listProps: DataSourceListProps;

  selectAll: {
    value: boolean;
    onValueChange: (isChecked: boolean) => void;
    indeterminate: boolean;
    isDisabled?: undefined;
  } | {
    value: boolean;
    onValueChange: () => void;
    isDisabled: boolean;
    indeterminate: boolean;
  },
}
```

This is a return type of `useDataRows` hook. Sometimes, all the properties, returned from this hook, are required. And not all of them should be passed to the `DataTable`. If listProps is spread too, it will turn into a mess. Don't you think that the best idea is to leave it as a separate property, which collects all the stats about the datasource.


Otherwise, we'll face the further issue:
```ts
const {
    getSelectedRows,
    getSelectedRowsCount,
    getById,
    clearAllChecked,
    selectAll,
    ...tableProps,
} = useDataRows(props);

<DataTable { ...tableProps } />;
```
To pass necessary data to a DataTable it will be required to destruct all the other props, which looks weird.
It seems to me, it is much better to extract only rows and listProps:
```ts
const { rows, listProps } = useDataRows(props);

<DataTable rows={ rows } { ...listProps } />;
```

@jakobz, WDYT about it?

uui-core/src/hooks/usePrevious.ts
```diff
@@ -1,8 +1,8 @@
 import { useEffect, useRef } from 'react';
 import isEqual from 'lodash.isequal';
 
-export function usePrevious<T>(value: T) {
-    const previousValueRef = useRef(value);
+export function usePrevious<T>(value: T): T | null {
```
jakobz, line 4: We currently export these hooks. We need to either remove them from exports for now, or document them. 

It's also not clear from their names what's the difference between them. Probably, it's better to have a single hook with a prop like 'compare'.

uui-core/src/data/processing/views/tree/treeStructure/helpers/PatchHelper.ts
```diff
@@ -0,0 +1,192 @@
...
+        patchIds: TId[],
+        originalIds: TId[],
+        {
+            getItemTemporaryOrder,
+            patchedItemsMap,
+            isDeleted,
+            complexIds,
+        }: ApplyPatchTemporaryReorderingOptions<TItem, TId>,
+    ) {
+        const tempOrderComparator = new Intl.Collator(undefined, { numeric: true, sensitivity: 'base' }).compare;
```
jakobz, line 76: We should use a simple, non Intl sting comparator here, to compare orders. It's faster and don't have issues with digits.

jakobz (commented):
Minor suggestions

uui-core/src/data/processing/views/tree/helpers/comparators.ts
```diff
@@ -0,0 +1,55 @@
...
+            const sortByFn = options.sortBy || ((i: TItem) => i[sortingOption.field as keyof TItem] ?? '');
+            const sign = sortingOption.direction === 'desc' ? -1 : 1;
+            comparators.push((a, b) => sign * compareScalars(sortByFn(a, sortingOption) + '', sortByFn(b, sortingOption) + ''));
+        });
+    }
+
+    return comparators;
+};
+
+export const composeComparetors = <TItem>(comparators: ((a: TItem, b: TItem) => number)[]) => {
```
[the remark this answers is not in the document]
Kuznietsov: Done.

uui-core/src/hooks/usePrevious.ts
Kuznietsov: Removed from exports, merged with useSimplePrevious, and added options with compare: boolean.

uui-core/src/data/processing/views/tree/treeStructure/helpers/PatchHelper.ts
Kuznietsov: Done.

uui-core/src/data/processing/views/tree/treeStructure/helpers/PatchHelper.ts
```diff
@@ -0,0 +1,200 @@
...
+        patchIds: TId[],
+        originalIds: TId[],
+        {
+            getItemTemporaryOrder,
+            patchedItemsMap,
+            isDeleted,
+            complexIds,
+        }: ApplyPatchTemporaryReorderingOptions<TItem, TId>,
+    ) {
+        const tempOrderComparator = (a: string, b: string) => a < b ? -1 : 1;
```
Kuznietsov, line 76: String comparators should be moved to the separate `stringComparator`.

Kuznietsov: Add `stringComparator` to `composeComparators`.

NatalliaAlieva:
Released in 5.8.0 ver.

Issue #1789, opened 2023-11-29
[PickerInput]: picker is automatically closed when unselecting value if the "show only selected" option is enabled


NatalliaAlieva:
## Description
The picker input automatically closes when User unselect any value if the "show only selected" option is enabled.

## Steps to Reproduce
1. Open Property Explorer for [PickerInput](https://uui.epam.com/documents?category=components&id=pickerInput&mode=propsEditor&skin=UUI) component.
2. Set `selectionMode = multi`.
3. Move to Picker input example.
4. Select several values.
5. Enable "Show only selected" option.
6. Unselect any value.

## Actual result
The picker input is automatically closed.

https://github.com/epam/UUI/assets/57751283/2923e644-1feb-4299-b8d8-103ffbc9364e

## Expected result
Picker input remains open.

## Environment

* UUI version: 5.4.0
* Browser name: any 
* Operating System: any

NatalliaAlieva:
Released in 5.5.0 ver.

Issue #1385, opened 2023-05-31
Input stays focused in case of outside click


stanislavec:
## Prerequisites

Please answer the following questions for yourself before submitting an issue. **YOU MAY DELETE THE PREREQUISITES SECTION.**

- [x] I am running the latest version
- [x] I checked the documentation and found no answer
- [x] I checked to make sure that this issue has not already been filed

## Description

Click outside input field doen't cause blur event on field in case if field is inside Dropdown.

## Steps to Reproduce

Try to wrap input fields with Dropdown container. So, fields could be inside Dropdown.

## Actual result

What is the current behavior?

[EPM_UUI_Dropdown_bug.webm](https://github.com/epam/UUI/assets/34837948/6056661b-6edd-4d2c-a186-da13b85078b7)

As I checked, the reason may be `isInteractedOutsideDropdown` function ([source code](https://github.com/epam/UUI/blob/main/uui-components/src/overlays/Dropdown.tsx)). 

## Expected result

Click outside input causes Blur event, so field starts to be unfocused.  

## Environment

* UUI version: 4.10.2
* Browser name: Chrome
* Operating System: MacOS

vik753:
My investigation found that after the 'clickOutside' event, the React remading DOM faster than the 'onClick' event is bubbling to target DOM Node. We can fix it to add timeout into the Dropdown/clickOutsideHandler method, but it's not the best decision. We will investigate this problem one more time in nearest future.

NatalliaAlieva:
Released in 5.5.0 ver.

Issue #1045, opened 2023-02-15
[Picker Input] Search field is not cleared, but simply closed when the search is inside body


NatalliaAlieva:
## Description
Search field is not cleared, but simply closed when the search is inside the body for singe and multiselect picker input.

## Steps to Reproduce
1. Open [Picker Input](https://uui.epam.com/documents?category=components&id=pickerInput&mode=propsEditor&skin=UUI4_promo) Property Explorer.
2. Set `searchPosition` = `body` OR `selectionMode` = `multi`.
3. Click on Picker Input demo.
4. Type any non-existent word in the search field.
5. Click on [X] icon in the search field.

## Actual result
Picker input closes.

https://user-images.githubusercontent.com/57751283/219048935-7175d65f-ddd6-4782-92f4-69b71724da3a.mp4

## Expected result
Search field is cleared, Picker input remains open.

## Environment
* UUI version: 4.10.0
* Browser name: any
* Operating System: any

NatalliaAlieva:
Released in 5.0.0 ver.

PR #1882, opened 2023-12-22
[Dropdown]: rework isInteractedOutsideDropdown helper to work with e.composedPath()

AlekseyManetov:
…composedPath(), rather than go through dom tree to detect if click was triggered on node inside parent stop nodes. It was an issues, that dom was updated earlier than event bubble to the parent node.

#### Issue link(if exists): #1789, #1385

**Testing notes:**
Need to retest:
- dropdown close on click outside logic
- Dropdown shouldn't close by clicks inside its body
- retest #1045 issue, since prior fix was reverted in flavor of this more versatile fix


NatalliaAlieva:
Released in 5.5.0 ver.

PR #1887, opened 2023-12-26
[SlateEditor] Fix colspan / rowspan & add merged cells to example

dimaanj:
## Summary

- Consider colSpan & rowSpan from the old `data` object
- Adjust table example with merged cells


NatalliaAlieva:
Released in 5.5.0 ver.

PR #1899, opened 2024-01-03
[NumericInput] Improve placeholder

### LINKED ISSUE #1860: [NumericInput]: placeholder doesn't appear after removing value if min>0
## Description

If User sets that the minimum value in the numeric input is greater than 0, then after deleting entered value, the minimum set value is overwritten, instead of the placeholder.

## Steps to Reproduce

1. Open [NumericInput](https://uui.epam.com/documents?category=components&id=numericInput&mode=propsEditor&isSkin=true&theme=loveship) component.
2. Set property min=10.
3. Input any value in the numeric input example.
4. Delete the entered value.
5. Press `delete` button.

## Actual result

The set minimum value appears instead of the placeholder.

https://github.com/epam/UUI/assets/57751283/e0a08a28-c10d-4afe-bdc4-9950f232e885

## Expected result

Placeholder appears.

## Environment

* UUI version: 5.4.2
* Browser name: any
* Operating System: any

dimaanj:
## Summary

- Fixes #1860
- Improves null checks

### DIFF
--- a/uui-components/src/inputs/NumericInput.tsx
+++ b/uui-components/src/inputs/NumericInput.tsx
@@ -92,7 +92,7 @@ export const NumericInput = React.forwardRef<HTMLDivElement, NumericInputProps>(
     const formatOptions = initialFormatOptions ?? { maximumFractionDigits: 0 };
 
     const placeholderValue = React.useMemo(() => {
-        if (!value && value !== 0) return props.placeholder || '0';
+        if (!value) return props.placeholder || '0';
         if (formatValue) return formatValue(value);
 
         return props.disableLocaleFormatting ? value.toString() : getSeparatedValue(value, formatOptions, i18n.locale);
--- a/uui-components/src/inputs/NumericInput.tsx
+++ b/uui-components/src/inputs/NumericInput.tsx
@@ -77,19 +77,28 @@ const getFractionDigits = (formatOptions: Intl.NumberFormatOptions) => {
 };
 
 export const NumericInput = React.forwardRef<HTMLDivElement, NumericInputProps>((props, ref) => {
-    let {
-        value = null, min, max, formatOptions,
+    const {
+        value: initialValue = null,
+        min: initialMin,
+        max: initialMax,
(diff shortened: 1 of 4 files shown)


uui-components/src/inputs/NumericInput.tsx
```diff
@@ -77,19 +77,28 @@ const getFractionDigits = (formatOptions: Intl.NumberFormatOptions) => {
...
+    const value = initialValue != null ? +initialValue : null;
+    const min = initialMin ?? 0;
+    const max = initialMax ?? Number.MAX_SAFE_INTEGER;
+    const formatOptions = initialFormatOptions ?? { maximumFractionDigits: 0 };
 
-    if (value != null) {
-        value = +value;
-    }
+    const placeholderValue = React.useMemo(() => {
+        if (!value && value !== 0) return props.placeholder || '0';
```
AlekseyManetov, line 95: Probably we can rewrite this condition to more straightforward: `if (value === null || value === undefined)`?

uui-core/src/helpers/__tests__/numericInputCalculations.test.tsx
```diff
@@ -29,7 +35,6 @@ describe('numericInputCalculations', () => {
         expect(getMinMaxValidatedValue({ value: 10, min: 0, max: 10 })).toBe(10);
         expect(getMinMaxValidatedValue({ value: 12, min: 0, max: 10 })).toBe(10);
         expect(getMinMaxValidatedValue({ value: -1, max: 10 })).toBe(0);
-        expect(getMinMaxValidatedValue({ value: null, max: 10 })).toBe(null);
```
AlekseyManetov, line 32: why do we remove this case?

AlekseyManetov (changes requested):

uui-core/src/helpers/__tests__/numericInputCalculations.test.tsx
dimaanj: ValidationDTO interface accept only `value: number`.

PR #1929, opened 2024-01-15
[NumericInput]: fixed sizes type (removed 'none')

vik753:


uui/components/inputs/NumericInput.tsx
```diff
@@ -10,11 +10,11 @@ const DEFAULT_MODE = EditMode.FORM;
 
 type NumericInputMods = IHasEditMode & {
     /**
-     * Defines component size.
+     * Defines component size
      * @default '36'
      * Size '48' is deprecated and will be removed in future release
      */
-    size?: ControlSize;
```
AlekseyManetov, line 17: Do we need `ControlSize` size at all? Where we are using it? Probably we can remove it?

AlekseyManetov (changes requested):

uui/components/inputs/NumericInput.tsx
vik753: removed 'none' from ControlSize

PR #1946, opened 2024-01-25
[Clickable]: Fixed problems found during testing

siarhei-epam:


uui-components/src/buttons/Button.tsx
```diff
@@ -1,10 +1,35 @@
 import * as React from 'react';
-import { Icon, uuiElement, uuiMarkers, CX, cx, IHasIcon, IDropdownToggler, IHasCaption, devLogger } from '@epam/uui-core';
-import { Clickable, ClickableComponentProps, UnionRawProps } from '../widgets';
+import {
```
AlekseyManetov, line 2: align imports according to our style-guides

uui-components/src/buttons/Button.tsx
```diff
@@ -16,20 +41,24 @@ export type ButtonProps = ClickableComponentProps & IDropdownToggler & IHasIcon
...
-export const Button = React.forwardRef<HTMLButtonElement | HTMLAnchorElement | HTMLSpanElement, ButtonProps>((props, ref) => {
+export const Button = React.forwardRef<HTMLButtonElement | HTMLAnchorElement, ButtonProps>((props, ref) => {
     if (__DEV__ && props.captionCX) {
         devLogger.warn('Button: Property \'captionCX\' is deprecated and will be removed in the future release. Please use \'cx\' prop to access caption styles and use cascading to change the styles for the \'uui-caption\' global class');
     }
 
     return (
         <Clickable
             { ...props }
+            href={ props.href }
```
AlekseyManetov, line 56: why do we pass  href separately, not via {...props}?

uui-components/src/buttons/Button.tsx
```diff
@@ -1,10 +1,35 @@
...
+    HrefNavigationProps,
+    LinkButtonNavigationProps,
+} from '../widgets';
 import { IconContainer } from '../layout';
 import css from './Button.module.scss';
 
-export type ButtonProps = ClickableComponentProps & IDropdownToggler & IHasIcon & IHasCaption & {
+export type UnionButtonNavigationProps = HrefNavigationProps | LinkButtonNavigationProps | ButtonNavigationProps | AnchorNavigationProps;
+
+export type ButtonRawProps =
```
AlekseyManetov, line 26: Can we use IHasRawProps<(React.AnchorHTMLAttributes<HTMLAnchorElement> | React.ButtonHTMLAttributes<HTMLButtonElement>> instead  of re-declaring the whole type here?

uui-components/src/buttons/IconButton.tsx
```diff
@@ -1,19 +1,37 @@
...
-export type IconButtonProps = ClickableComponentProps & IDropdownToggler & {
+export type UnionIconButtonNavigationProps = HrefNavigationProps | LinkButtonNavigationProps | ButtonNavigationProps | AnchorNavigationProps;
+
+export type IconButtonRawProps =
+    | (React.AnchorHTMLAttributes<HTMLAnchorElement>
+    | React.ButtonHTMLAttributes<HTMLButtonElement>)
+    & Record<`data-${string}`, string>;
+
+export type IconButtonProps = IClickable & IAnalyticableClick & IHasTabIndex & IDisableable & IHasCX
+& Omit<ICanRedirect, 'href' | 'link'> & IDropdownToggler & UnionIconButtonNavigationProps & {
```
AlekseyManetov, line 15: why do we omit here 'href' and 'link'?

uui-components/src/widgets/Clickable.tsx
```diff
@@ -1,64 +1,68 @@
...
+    rawProps?: ClickableRawProps;
+};
 
 export const Clickable = React.forwardRef<HTMLButtonElement | HTMLAnchorElement | HTMLSpanElement, PropsWithChildren<ClickableComponentProps>>((props, ref) => {
     const context = useUuiContext();
     const isAnchor = Boolean(props.href || props.link);
-    const isButton = Boolean(!isAnchor && props.onClick);
-    const isClickable = Boolean(!props.isDisabled && (isAnchor || props.onClick));
+    const isButton = Boolean(!isAnchor && ((props.rawProps as React.ButtonHTMLAttributes<HTMLButtonElement>)?.type || props.onClick));
+    const isClickable = Boolean(!props.isDisabled && (props.link || props.onClick));
```
AlekseyManetov, line 65: Why do we not include props.href in this condition?

uui-components/src/widgets/Clickable.tsx
```diff
@@ -74,6 +78,11 @@ export const Clickable = React.forwardRef<HTMLButtonElement | HTMLAnchorElement
             }
 
             if (!!props.link) {
+                if (props.target) {
```
AlekseyManetov, line 81: For which case do we need it? It's a danger to add e.stopPropagation() component, because we are not sure that user don't have case to also catch this event somewhere else.

AlekseyManetov (changes requested):

uui-components/src/widgets/Clickable.tsx
siarhei-epam: We need it to add possibility open links in new tab using target = "_blank"

uui-components/src/widgets/Clickable.tsx
siarhei-epam: It uses to pass click handler to components for "link" & "onClick" props, I will rename "isClickable" to "hasClick"

uui-components/src/buttons/IconButton.tsx
siarhei-epam: We have discriminated union, in my point of view in this case - 'href' | 'link' redundant

uui-components/src/buttons/Button.tsx
siarhei-epam: I will try to change

uui-components/src/buttons/Button.tsx
siarhei-epam: Typescript can't resolve href using spread

uui-components/src/buttons/IconButton.tsx
AlekseyManetov: but why we shouldn't do the same for all other components?

PR #2079, opened 2024-03-19
Filecard component i18n translation

alekse1ka:
### Description:
Not all strings in the FileCard component moved to the translation file. This will give more flexibility for multilanguage websites that will need to translate everything.


AlekseyManetov:
@alekse1ka Thanks for your contribution!

PR #2094, opened 2024-03-29
Use tree simpler temp order handling

jakobz:
### Description:

Simpler tempOrder handling in Sandbox / Editable Table


Kuznietsov (approved):
LGTM 👍

PR #2117, opened 2024-04-04
[UUICore]: cleaned up packages and helpers from uui-core.

Kuznietsov:
### Summary

- [uui-core]: helpers cleanup
  * Deprecated:
    * `LazyLoadedMap` and related:
      * `LazyLoadedMapLoadCallback`
      * `UNKNOWN`
      * `LOADING`
      * `LOADED`
      * `PENDING`
      * `FAILED`
      * `LoadingStatus`
    * `browser` helper:
      * `Browser`
      * `getBrowser`
    * `Debouncer`
    * `parseIconViewbox`
    * `parseStringToCSSProperties`
    * `getScreenSize`
    * `urlParser`
    * `batch` and related:
      * `batch`
      * `BatchPromiseOptions`

-  Removed unused packages.


uui-core/src/helpers/range.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    if (start < end) {
+        yield* rangeFromTo(start, end, step);
+    } else {
+        yield* reverseRangeFromTo(start, end, step);
+    }
+}
+
+export const range = (start: number, end?: number, step: number = 1) =>
+    Array.from(rangeGen(start, end, step));
```
jakobz, line 26: Do we actually need to support cases except when start < end, and step = 1? We are not forced to have an lodash-compatible drop-in replacement. We need only to cover our cases.

jakobz, line 26: Can't we return iterators from 'range' function, w/o converting them to array?

JS-Iterator-based API for 'range' can be a good idea, as in this case we don't create an array just to iterate over it 
indexes and throw it away.

If we return array anyway, it will be simpler and faster to go old-school like:

const result = new Array(end - start); // to allocate space for items
for(var n = start; n < end; n++) arr[n] = n;

uui-core/src/helpers/orderBy.ts
```diff
@@ -0,0 +1,17 @@
...
+    if (typeof by === 'function') {
+        const sign = order === 'desc' ? -1 : 1;
+        return sortedArr.sort((a, b) => {
+            const aBy = by(a);
+            const bBy = by(b);
+            return compareScalars(aBy, bBy, sign);
+        });
+    }
+
+    const comparer = getOrderComparer([{ field: by, direction: order }]);
```
jakobz, line 15: We use a simple string comparator if `by` is a function. However, we use Intl-based comparator (implied by use of getOrderComparer) - if we pass a field name as a string.

Why do we need an overloads here? Probably we can accept a function only, and rewrite any `orderBy(arr, 'name')` to orderBy(arr, i => i.name)?

uui-core/src/helpers/range.ts
Kuznietsov: The previous version returned a generator only, but all over the project `.map/.forEach` Array methods are used.
That's why I've returned an array from generator.
But, probably, we can make it simpler by generating array.

jakobz (approved):

PR #2137, opened 2024-04-10
[NextApp] Introduce next app router example

dimaanj:
## Summary

Introduce next app router example


templates/uui-nextjs-template-app-router/components/AppView.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Modals } from "@epam/uui-components";
+
+export async function AppView({
+    children,
+    appContext,
+}: PropsWithChildren<{ appContext: AppContextType }>) {
+    const { services } = useUuiServices<TApi, AppContextType>({
+        appContext,
+        apiDefinition,
+        router: new StubAdaptedRouter(),
```
AlekseyManetov, line 23: Let's use here adapted to UUI router

templates/uui-nextjs-template-app-router/helpers/appContext.ts
```diff
@@ -0,0 +1,8 @@
+export async function getAppContext() {
```
AlekseyManetov, line 1: Let's remove this helper and usage since using of App context is not popular case

templates/uui-nextjs-template-app-router/helpers/apiDefinition.ts
```diff
@@ -0,0 +1,15 @@
+import { IProcessRequest } from "@epam/uui-core";
+
+export function apiDefinition(processRequest: IProcessRequest) {
```
AlekseyManetov, line 3: Let's remove this helper and usage since using of it isn't required and current implementation does not provide any benefit for user

AlekseyManetov (changes requested):

Issue #1822, opened 2023-12-07
[PIckerInput]: entered value clears when user put pointer in input repeatedly


mari-shakhanava:
## Description (optional)

This issue is about disappearing entered value when repeated putting pointer in the input field.

 ## Steps to Reproduce

1. PickerInput with searchPosition='input' prop.
2. Start typing anything
3. After appearing values in dropdown put pointer back in the input

https://github.com/epam/UUI/assets/70166849/8bbb4d45-a2c5-4edf-9e03-167da7d6fb26

## Actual result

Entered value cleared

## Expected result

The entered text should stay unchanged. As is in PickerInput with searchPosition='body', so user can continue his search

## Environment

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* UUI version: latest
* Browser name: Chrome latest
* Operating System: Windows latest

AlekseyManetov:
Such behavior isn't related to the search typing, it's happening because we are clearing search value by picker close and in your case click in toggler of already opened Picker leads to its closing. I'm not sure that can do something with it, because both features - 'Toggle opened state by toggler click' and 'Clear search after dropdown close' is an a core ux requirements of this component. Probably we need to discuss, whether we need to close picker by click in toggler? @DzmitryTrubchyk what do you think?
As a one more solution, I will recommend to use `searchPosition='body'` for majority of cases, because this pattern more clear and not so confused according to the UX.

AlekseyManetov:
We've decided that we should not close body by click on toggler in case when searchPosition='input' and body is open now.

DzmitryTrubchyk:
[like]  Dzmitry Trubchyk reacted to your message:
________________________________

PR #2157, opened 2024-04-12
Bugfix/PickerInput don't close body #1822

vik753:


uui-components/src/pickers/PickerToggler.tsx
```diff
@@ -199,11 +199,16 @@ function PickerTogglerComponent<TItem, TId>(props: PickerTogglerProps<TItem, TId
         );
     };
 
+    const shouldToggleBody = (e: React.MouseEvent<HTMLDivElement>): boolean => {
+        const isInteractionDisabled = (props.isDisabled || props.isReadonly || isEventTargetInsideClickable(e));
+        const isInputFocusedAndValid = (inFocus && props.value && props.minCharsToSearch);
```
AlekseyManetov, line 204: Let's better rename to 'shouldOpenWithMinCharsToSearch'

uui-components/src/pickers/PickerToggler.tsx
```diff
@@ -199,11 +199,16 @@ function PickerTogglerComponent<TItem, TId>(props: PickerTogglerProps<TItem, TId
...
+        const isPickerOpenForInput = (props.isOpen && props.searchPosition === 'input');
```
AlekseyManetov, line 205: Let's better rename to 'isPickerOpenWithSearchInInput' and add comment below 'We shouldn't close the picker by click on input in toggler, because it will lead to the search clearing by body close'.

uui-components/src/pickers/PickerToggler.tsx
vik753: done

uui-components/src/pickers/PickerToggler.tsx
vik753: done

Issue #2159, opened 2024-04-13
Blocker - TransitionGroup-related issues, e.g. flaky PickerInput test


jakobz:
## Now

We have Blocker component, which use TransitionGroup to trigger animations by adding/removing elements to the DOM. Animation speed is controlled with CSS. However, how long we keep Blocker in the DOM, is controlled by TransitionGroup. It's ok, as after animation is finished, we can keep a node in as long as needed - it's not visible anyway.

However, there are some issues:

- PickerInput tests waits for Blocker to disappear from DOM, and both waitFor and TransitionGroup timeouts are 1000ms. This makes tests flaky. Just tweaking timeouts would work. However, in real life you can use Picker immediately after Blocker gets isEnabled=false. When got isEnabled=false, it immediately gets pointer-events: none, so you can click through it. Tests shouldn't wait for animation to finish either.
- a18y-wise, we also want DOM state to reflect that Blocker is disabled. 

# Suggested solution

- [Stack overflow](https://stackoverflow.com/questions/38704467/how-to-label-a-loading-animation-for-wai-aria) suggest to use aria-busy=true|false for this. Let's use it. This way aria-busy would reflect Blocker state in DOM, w/o any animation delays

- Let's update our tests to use aria-busy to detect if loading is done. This would remove the need to wait for animation and TransitionGroup to complete.

- Let's also replace 'alert' role with 'presentation' or 'none'. The Blocker is a visual component, and doesn't add any semantic. Check that we have aria-busy role somewhere on the picker/table rows containers to indicate that they are loading.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/presentation_role

PR #2164, opened 2024-04-15
[Pickers]: Blocker role/aria-label/aria-busy added.

Kuznietsov:
### Summary
Closing: https://github.com/epam/UUI/issues/2159

 
- Blocker role/aria-label/aria-busy have changed
- Tests updated.

### DIFF
--- a/uui/components/layout/VirtualList.tsx
+++ b/uui/components/layout/VirtualList.tsx
@@ -118,15 +118,7 @@ const VirtualListView = React.forwardRef<HTMLDivElement, VirtualListViewProps>((
             >
                 { props.children }
             </div>
-            <Blocker
-                isEnabled={ props.isLoading }
-                rawProps={ { 
-                    role: 'status',
-                    'aria-live': 'polite',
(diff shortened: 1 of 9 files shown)


uui/components/layout/VirtualList.tsx
```diff
@@ -118,7 +118,15 @@ const VirtualListView = React.forwardRef<HTMLDivElement, VirtualListViewProps>((
...
             </div>
-            <Blocker isEnabled={ props.isLoading } />
+            <Blocker
+                isEnabled={ props.isLoading }
+                rawProps={ { 
+                    role: 'status',
+                    'aria-live': 'polite',
+                    'aria-busy': props.isLoading ? 'true' : 'false',
+                    'aria-label': props.isLoading ? 'Loading' : '',
+                } }
```
jakobz, line 128: Can we put these aria attributes directly into the Blocker component? We can use 'isEnabled' there, the same way we use isLoading here.

jakobz (approved):

PR #2172, opened 2024-04-17
[Pickers]: fixed working with dataSourceState updates.

Kuznietsov:
## Summary

Fixed working with dataSourceState updates and all bugs, related to dataSourceState pinning and warnings.

### DIFF
--- a/uui/components/filters/FilterPickerBody.tsx
+++ b/uui/components/filters/FilterPickerBody.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { DataRowProps, DataSourceListProps, DropdownBodyProps, isMobile, PickerFilterConfig } from '@epam/uui-core';
+import { DataRowProps, DataSourceListProps, DropdownBodyProps, isMobile, PickerFilterConfig, usePrevious } from '@epam/uui-core';
 import { PickerBodyBaseProps, PickerInputBaseProps, usePickerInput } from '@epam/uui-components';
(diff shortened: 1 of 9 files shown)


uui/components/filters/FilterPickerBody.tsx
```diff
@@ -23,13 +23,24 @@ export function FilterPickerBody<TItem, TId>(props: FilterPickerBodyProps<TItem,
...
+    const prevOpened = usePrevious(props.isOpen);
+
     const renderItem = (item: TItem, rowProps: DataRowProps<TItem, TId>) => {
         return <PickerItem title={ getName(item) } size="36" { ...rowProps } />;
     };
 
+    React.useLayoutEffect(() => {
+        if (prevOpened === props.isOpen && props.isOpen 
+            && prevValue !== props.value && props.value !== props.emptyValue
+            && props.selectionMode === 'single'
```
Kuznietsov, line 36: A body was closing previously in `single` selection mode on filter value selection.
On the other hand, if `clear` was executed, a body was not closed. To support such behavior, logic `props.value !== props.emptyValue` was added.

@AlekseyManetov, WDYT, is it correct, or body should be closed on both actions, filter value select and filter value clear?

AlekseyManetov: Yes, we shouldn't close picker the body by 'Clear' actions, because it's a common use case when user open picker -> clear value -> select another.

PR #2232, opened 2024-05-08
[Plate Editor] Enhance content migration strategies

### LINKED ISSUE #1961: [RTE]: add html serialization for separatorPlugin
## Description
Let's add support of HTML serialization for the `separatorPlugin`. We can serialize separator to the native `<hr />` html tag.

dimaanj:
## Summary

- enhance content migration strategy
- made `strictNullCheck` ready
- `value` prop update from outside now works faster than remounting via react `keys`
- get rid of element mutations
- get rid of circular deps
- placeholder now renders simultaneously
- improve plugin imports, types and config
- resolves #1961


uui-editor/src/plugins/colorPlugin/colorPlugin.tsx
```diff
@@ -33,8 +32,10 @@ interface IToolbarButton {
     editor: PlateEditor;
 }
 
+const modifiers = [{ name: 'offset', options: { offset: [0, 3] } }];
```
AlekseyManetov, line 35: Let's move this global variable into the component, because global variable assignment can be assumed as a side effect for bundlers, we should avoid using global scope variables, if possible

AlekseyManetov (changes requested):

app/src/demo/RTE/RichTextEditorDemo.tsx
```diff
@@ -47,7 +47,7 @@ export class RichTextEditorDemo extends React.Component<any, SlateEditorBasicExa
 
     plugins = [
         ...defaultPlugins,
-        baseMarksPlugin(),
+        ...baseMarksPlugin(),
```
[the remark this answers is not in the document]
dimaanj: It won't be a breaking change. We forgot to flatten base marks plugins array some time ago, but we do `plugins.flat()` anyway in `SlateEditor`

uui-editor/src/plugins/colorPlugin/colorPlugin.tsx
dimaanj: It's not a global vaiable, it's inside the js module. I did it to avoid redundant use of `useMemo` to enable memoization

AlekseyManetov: yes, I understand, but for bundler each variable which created in the global scope(not inside the function or class) assumed like a side effect and if we have such things, we can't mark or package as `sideEffect: false`. But anyway Dropdown isn't wrapped in React.Memo, so it doesn't have any sense to wrap it into `useMemo` and also you still have inline arrow function in renderTarget prop, so this change doesn't make any optimizations ;)

uui-editor/src/plugins/uploadFilePlugin/file_uploader.ts
```diff
@@ -24,22 +27,25 @@ type UploadFile = (
...
     }),
-    image: (file: FileUploadResponse) => ({
-        type: IMAGE_PLUGIN_TYPE,
-        data: file,
-        url: file.path,
+    image: (file: FileUploadResponse): TImageElement => ({
+        type: IMAGE_TYPE,
+        data: {
+            ...file, // TODO: remove, seems irrelevant
+            size: {},
```
AlekseyManetov, line 41: why do we replace size filed to empty object? Also, comments about removing file from data looks outdated regarding our previous sync up.

uui-editor/src/plugins/videoPlugin/AddVideoModal.tsx
```diff
@@ -64,7 +64,15 @@ export function AddVideoModal({ editor, success, abort, ...props }: AddVideoModa
                 <ModalHeader title="Add video" onClose={ abort } />
                 <FlexRow cx={ css.inputWrapper }>
                     <LabeledInput label="Video url">
-                        <TextInput value={ src } onValueChange={ setSrc } autoFocus />
+                        <TextInput
+                            value={ src }
+                            onValueChange={ (v) => {
+                                if (v) {
```
AlekseyManetov, line 70: Why do we need such condition here? I think it will broke set of empty value to the input, in case when you have some value and then decided to remove it.

AlekseyManetov (changes requested):

uui-editor/src/plugins/colorPlugin/colorPlugin.tsx
dimaanj: From my view nothing wrong with that code

app/src/demo/RTE/RichTextEditorDemo.tsx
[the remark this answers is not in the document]
dimaanj: I've updated type for plugins from `any` to `PlatePlugin[]` and removed redundant flattening. So, from now we have to pass spreaded `...baseMarksPlugin()`

uui-editor/src/plugins/uploadFilePlugin/file_uploader.ts
dimaanj: The fields seems to completely irrelevant except `path` for iframe / image and `name` of the file for attachment. Other fields should be skipped to keep content clean.

uui-editor/src/plugins/videoPlugin/AddVideoModal.tsx
dimaanj: Changed

uui-editor/src/migrations/plate_migrations.ts
```diff
@@ -0,0 +1,69 @@
+import { Value, setNodes, PlateEditor, TNodeEntry } from '@udecode/plate-common';
```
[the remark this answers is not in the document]
dimaanj: Renamed

uui-editor/src/__tests__/slate-migration.test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+describe('migrate', () => {
+    describe('slate', () => {
+        it('should migrate content correctly', () => {
+            const migrated = migrateSlateSchemaNew(slateInitialValue);
+
+            expect(migrated).toMatchSnapshot();
+        });
+
+        it('should migrate full slate content', () => {
```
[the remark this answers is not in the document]
dimaanj: Removed short content

uui-editor/src/plugins/uploadFilePlugin/file_uploader.ts
dimaanj: Discussed offline

uui-editor/src/plugins/colorPlugin/colorPlugin.tsx
dimaanj: Discussed offline

app/src/demo/RTE/RichTextEditorDemo.tsx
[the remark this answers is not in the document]
dimaanj: Discussed offline

Issue #2198, opened 2024-04-23
[DataTable]: condensed view


AlekseyManetov:
## To Do

- Add `headerSize: '36' | '48'` prop. Default '36'
- In `headerSize='48'` we can have 2 line header, when text wrapping
- Add `columnsGap: '12' | '24'` prop, default '24'. The prop will configure space between columns. Add mention that '12' value, will not work with editable tables.
- Add condensed table example with DataTable headerSize='36', columnsGap: '12' and size='30'(row size), like in design below.

Design - https://www.figma.com/file/cAd3iaIW4JxJapIVTrE96b/UUI-%E2%80%93-Tables-%26-Filters-(Demo-%26-Examples)?type=design&node-id=6321-28475&mode=design&t=Cp0I0rD222k0MXFJ-0

NatalliaAlieva:
Released in 5.8.0 ver.

PR #2261, opened 2024-05-20
Feature/DataTable condensed view #2198

vik753:


uui-core/src/types/tables.ts
```diff
@@ -147,6 +147,8 @@ export interface DataTableHeaderCellProps<TItem = any, TId = any> extends IEdita
     onSort(dir: SortDirection): void;
     onDrop?(params: DropParams<DataColumnProps<TItem, TId>, DataColumnProps<TItem, TId>>): void;
     renderFilter?: (dropdownProps: IDropdownBodyProps) => React.ReactNode;
+    /** Defines table columns gap size */
+    columnsGap?: '12' | '24';
```
AlekseyManetov, line 151: Add this prop to the DataTableCellMods interface in uui package, since DataRowProps it's a theme agnostic core interface, which shouldn't include mods and sizes.
Also add mention about '@default' value(related to the other interfaces too)

uui-core/src/types/tables.ts
```diff
@@ -164,6 +166,8 @@ export interface DataTableHeaderRowProps<TItem = any, TId = any> extends IEditab
     onConfigButtonClick?: (params: DataTableConfigModalParams) => any;
     renderCell?: (props: DataTableHeaderCellProps<TItem, TId>) => React.ReactNode;
     renderConfigButton?: () => React.ReactNode;
+    /** Defines table columns gap size */
+    columnsGap?: '12' | '24';
```
AlekseyManetov, line 170: Why do we need columnsGap in column configuration modal?

uui-core/src/types/tables.ts
```diff
@@ -251,6 +255,9 @@ export interface DataTableCellProps<TItem = any, TId = any, TCellValue = any> ex
 
     /** Overrides default tooltip, used to show validation message if the cell is invalid */
     renderTooltip?: (props: ICanBeInvalid & TooltipCoreProps) => React.ReactElement;
+
+    /** Defines columns gap size */
+    columnsGap?: '12' | '24';
```
AlekseyManetov, line 260: I think, `columnsGap` is a global table prop, we shouldn't give a opportunity to define gap for each columns separately, so we can remove this prop from this interface.

AlekseyManetov (changes requested):

uui-core/src/types/dataRows.ts
```diff
@@ -107,6 +107,9 @@ DataRowOptions<TItem, TId> & {
 
     /** True if row pinned, it means that it will be sticky inside his nesting level */
     isPinned?: boolean;
+
+    /** Defines columns gap size */
+    columnsGap?: '12' | '24';
```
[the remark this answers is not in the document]
vik753: done

uui-core/src/types/tables.ts
vik753: done

uui-core/src/types/tables.ts
vik753: moved it to the uui/ DataTableHeaderRowMods

uui-core/src/types/tables.ts
vik753: done

PR #2334, opened 2024-06-07
Modals refactored to FC and tests

vik753:


NatalliaAlieva:
Released in 5.8.2 ver.

PR #2357, opened 2024-06-13
[SlateEditor] Add plugin cusomizations

dimaanj:
## Summary

- add headers customisations
- add color customsations
- add note customisations
- move components to local plugin folders


uui-editor/src/implementation/ToolbarButton.tsx
```diff
@@ -11,17 +11,25 @@ interface ToolbarButtonProps extends IHasCX, IHasCaption, IDisableable {
...
-        } }
-        icon={ props.icon }
-        caption={ props.caption }
-        ref={ ref }
-        cx={ cx(css.toolbarButton, css['color-' + props.iconColor], css[props.isActive ? 'gray90' : 'gray80'], props.cx) }
-        isDisabled={ props.isDisabled }
-    />
-));
+export const ToolbarButton = React.forwardRef<HTMLButtonElement | HTMLAnchorElement, ToolbarButtonProps>((props, ref) => {
+    const rawProps = React.useMemo(() => {
```
AlekseyManetov, line 18: useMemo is useless here, because we didn't make any calculations inside. Moreover using the useMemo for such cases will be less performance, than without it. I supposed we can just inline it.

uui-editor/src/plugins/colorPlugin/ColorBar.tsx
```diff
@@ -0,0 +1,77 @@
...
+};
+
+export function ColorBar({ onColorUpdate, onColorClear, value }: IColorBar) {
+    const editorRef = useEditorRef();
+    const { colors: colorProps } = React.useMemo(() => getPluginOptions<ColorPluginOptions>(
+        editorRef,
+        COLOR_PLUGIN_KEY,
+    ), [editorRef]);
+
+    const defaultColors = React.useMemo(() => (
```
AlekseyManetov, line 26: the same case with useless useMemo

uui-editor/src/plugins/colorPlugin/ColorBar.tsx
```diff
@@ -0,0 +1,77 @@
...
+                key="success"
+                onClick={ () => onColorUpdate('success') }
+                iconColor="green"
+                isActive={ value === 'success' }
+                icon={ ColorIcon }
+            />
+        </React.Fragment>
+    ), [onColorUpdate, value]);
+
+    const userColors = React.useMemo(() =>
```
AlekseyManetov, line 52: the same case with useless useMemo

AlekseyManetov (changes requested):

NatalliaAlieva:
Released in 5.8.2 ver.

PR #2359, opened 2024-06-14
[inputs]: removed pointer-events: none

vik753:
### Description: [inputs]: removed pointer-events: none
Fixed tooltip on '+N' `PickerInput` tag in disabled and readonly states.


NatalliaAlieva:
Released in 5.8.2 ver.

PR #2384, opened 2024-06-25
[SlateEditor] Fix table resize after paste

dimaanj:
## Summary

Fix table resize after paste from word


NatalliaAlieva:
Released in 5.8.2 ver.

PR #2411, opened 2024-07-12
Improve `ApiContext`

vasilii-kovalev:
### Description

* Combine user headers with custom ones in `startCall` method
* Assign `Content-Type` header only if it is not already present
* Add `ResponseData` generic type for `processRequest` function
* Add `ProcessRequestMethod` type
* Sync changelog with `main` branch
* Described changes in `changelog.md`
* Refactor tests for `ApiContext`:
  * Use [AAA pattern](https://uui.epam.com/documents?id=testing-basics&mode=doc&category=testing#:~:text=Recommended%20structure%20of%20a%20single%20test), separate groups with empty lines
  * Get rid of `any` types
  * Move `parseBlob` to a standalone function
  * Pass response data type to `processRequest` calls (except where `parseResponse` is defined)
  * Improve formatting to ease adding/removing properties/function parameters


app/src/data/apiDefinition.ts
```diff
@@ -33,27 +33,20 @@ export function getApi({
...
-        ...requestParams: Parameters<IProcessRequest<ResponseData>>
-    ) => {
-        const [
-            url,
-            method,
-            data,
-            options,
-        ] = requestParams;
-
+    const processRequestLocal: IProcessRequest = (url, method, data, options) => {
```
AlekseyManetov, line 36: Do we need `processRequestLocal` here? Probably we can use native processRequest method

AlekseyManetov (changes requested):

PR #2427, opened 2024-07-25
[DatePickers]: improve a11y focus flow behavior

AlekseyManetov:
### Description:
Now date picker body receive focus on open and return it back on input after close.


NatalliaAlieva:
Released in 5.9.0 ver.

PR #2451, opened 2024-08-06
[DataTable]: Added and reused DataTableCellContainer.

Kuznietsov:
### Summary

`DataTableCellContainer` was added and reused in `DataTableCell` and `DataTableHeaderCell`.


uui/components/tables/__tests__/__snapshots__/DataTableHeaderCell.test.tsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`DataTableHeaderCell should be rendered correctly 1`] = `
 <div
   aria-sort="none"
-  className="flexCell uui-dt-vars uui-table-header-cell cell size-36 first-column-24 pinned-left"
```
Kuznietsov, line 6: `flexCell` is removed, because `FlexCell` from `uui-components` is used now in `DataTableHeaderCell`.

@jakobz, @AlekseyManetov, WDYT? What should be done? `.flexCell` moved to `uui-components` or leave it as it is?

AlekseyManetov: it's ok, because this class was not a globa, so it can't be used in userland.

PR #2501, opened 2024-09-09
[PickerInput]: Added possibility to type symbols in search field for multiselect when minCharsToSearch > 0 and searchPosition in the body

Kuznietsov:
### Summary

In progress...


uui-core/src/types/pickers.ts
```diff
@@ -70,10 +82,32 @@ export type PickerBaseOptions<TItem, TId> = {
...
+     * */
     renderNotFound?: (props: { search: string; onClose: () => void }) => ReactNode;
 
+    /**
+     * Overrides the rendering of PickerInput body content when it is empty for various reasons.
+     * If provided, the override should support messages for both 'not-found-records' and 'less-than-min-chars-to-search' reasons.
+     * If not provided, the `renderNotFound` method is used for the 'not-found-records' reason.
+     * @param props
+     * @param props.reason - Specifies the reason why the PickerInput body is empty:
+     *   - 'not-found-records': No data was found.
```
jakobz, line 98: There are 3 cases when there are no records in the Picker Body:

1. There are no items at all, e.g. empty array is passed
2. Search was run, but returned no elements
3. There are maxCharsToSearch set, and not enough chars entered to trigger search

1 and 2 can be distinguished with search != ''. However, 2 and 3 can also be distinguished with search.length < minCharsToSearch.

So I feel like this API is not consistent - we have 2 reasons in `reason`, instead of 3. And there's no clear logic why we choose this set.

Also, 'found' in 'not-found-records' imply that we were searching something, while in case 1 we did not.

To make API consistent, I'd either 
A) add a third reason, e.g. 'no-data'
B) remove the reason prop altogether. We can add `isSearchTooShort: boolean` to avoid adding using complex `search.length < minCharsToSearch`.

While I voted for A before, now I'm more into B. Because who knows what other reasons we got later? It can be hard to classify and name them. For example, if we'll opt to handle errors, should we distinguish errors with and without search? E.e. should we have separate "search-error" and "load-error" reasons?

uui-core/src/types/pickers.ts
```diff
@@ -70,10 +82,32 @@ export type PickerBaseOptions<TItem, TId> = {
...
+     *   - 'less-than-min-chars-to-search': The search contains fewer characters than required to start searching for data.
```
jakobz, line 99: Let's rename "less-than-min-chars-to-search" to "search-too-short"

uui-core/src/types/pickers.ts
```diff
@@ -122,6 +156,10 @@ export type PickerFooterProps<TItem, TId> = {
     selection: PickerBindingProps<TItem, TId>['value'];
     /** Defines a search value */
     search: string;
+    /**
+     * Indicates whether the search does not contain enough characters to load data.
+     */
+    notEnoughTokensToLoadData?: boolean;
```
jakobz, line 162: isSearchTooShort: boolean ?

uui-core/src/types/pickers.ts
```diff
@@ -110,7 +144,7 @@ export type PickerBaseOptions<TItem, TId> = {
 export type PickerFooterProps<TItem, TId> = {
     /** Instance of picker DataSource view */
     view: IDataSourceView<TItem, TId, any>;
-    /** IEditable interface for the 'Show only selected' toggler */
+    /** IEditable interface for the 'ShowrenderTypeSearchToLoadItems only selected' toggler */
```
jakobz, line 147: 'ShowrenderTypeSearchToLoadItem' - looks odd...

uui/components/pickers/DataPickerBody.tsx
```diff
@@ -27,6 +28,8 @@ export interface DataPickerBodyProps extends Overwrite<DataPickerBodyMods, DataP
 export class DataPickerBody extends PickerBodyBase<DataPickerBodyProps> {
     lens = Lens.onEditableComponent<DataSourceState>(this);
     searchLens = this.lens.prop('search');
+    getSearchSize = () => (isMobile() ? settings.sizes.pickerInput.body.mobile.searchInput : this.props.searchSize) as SearchInputProps['size'];
+
     renderNotFound() {
```
jakobz, line 33: Let's mix this logic into renderEmpty.

uui/components/pickers/DataPickerBody.tsx
```diff
@@ -39,8 +42,27 @@ export class DataPickerBody extends PickerBodyBase<DataPickerBodyProps> {
         );
     }
 
+    renderEmpty(reason: PickerEmptyBodyReason) {
+        if (this.props.renderEmpty) {
+            return this.props.renderEmpty(reason);
+        }
+
+        if (reason === 'not-found-records') {
+            return this.renderNotFound();
```
jakobz, line 51: Don't we need a default implementation for renderNotFound somewere here?

Issue #2510, opened 2024-09-16
[DataTable] [LazyDataSource]: DataTable with LazyDataSource does not load items on scroll in certain cases. The component crashes when presets panels is present.


NatalliaAlieva:
### Description
It seems this bug is reproduced on any DataTable with LazyDataSource. E.g. here:
https://uui.epam.com/sandbox?id=productsTableDemo

### Prerequisites
Make sure the amount of rows fit on the screen is more or less big (like 30 rows or more). To achieve this, you can use browser zoom or chrome dev tools features to make logical screen height 1200px or more.

## Use case
Open Data Table example with Presets panel: https://uui.epam.com/docExample?theme=loveship&examplePath=tables%2FPresetsPanelBasic
Make sure screen heigh is big enough or use zoom out features of browser to achive this.
Open dev tools and set height: 100% style to the .uui-panel element to make sure it occupies all available height.
Actual result: The component crashes with the following error: https://react.dev/errors/185?invariant=185

#### Environment
Reproduced on any relevant example on https://uui.epam.com/ site

NatalliaAlieva:
Released in 5.10.0 ver.

PR #2511, opened 2024-09-16
[VirtualList]: Added logging of error for not limited VirtualList parent container.

Kuznietsov:
### Summary

Fixed https://github.com/epam/UUI/issues/2510

Logging of error for not limited VirtualList parent container was added.

Before, the component was crashing without any error.
Now, the component stops the infinite rows count increasing and logs an error about not limited container height.

### DIFF
--- a/uui-core/src/hooks/useVirtualList/useVirtualList.ts
+++ b/uui-core/src/hooks/useVirtualList/useVirtualList.ts
@@ -7,6 +7,7 @@ import {
 } from './utils';
 import { VirtualListInfo, UseVirtualListProps, UseVirtualListApi, RowsInfo } from './types';
 import { usePrevious } from '../usePrevious';
+import { devLogger } from '../../helpers';
 
 export function useVirtualList<List extends HTMLElement = any, ScrollContainer extends HTMLElement = any>(
     props: UseVirtualListProps,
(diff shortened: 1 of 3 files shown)


uui-core/src/hooks/useVirtualList/useVirtualList.ts
```diff
@@ -53,6 +57,20 @@ export function useVirtualList<List extends HTMLElement = any, ScrollContainer e
...
+        if (scrollContainer.current?.clientHeight
+            && prevScrollContainerClientHeight
+            && scrollContainer.current?.clientHeight !== prevScrollContainerClientHeight
+        ) {
+            ++scrollContainerHeightChangesCount.current;
+        }
+        if (scrollContainerHeightChangesCount.current > 20 && !scrollContainerHeightIsNotLimited.current) {
+            scrollContainerHeightIsNotLimited.current = true;
+            console.error(`The scroll container height is unlimited, likely because the parent container's height is set to 100%. Please ensure that the VirtualList's parent container has a defined, limited height.
+It is recommended to use viewport height (vh) units to set a relative height.`);
```
jakobz, line 70: Let's not guess for reasons here. It can be misleading. Just `The scroll container height is not limited. Please ensure that the VirtualList's parent container has a defined, limited height.` - will be good enough.

Also, it might be a good idea to provide a link to the VirtualList docs.

uui-core/src/hooks/useVirtualList/useVirtualList.ts
```diff
@@ -53,6 +57,20 @@ export function useVirtualList<List extends HTMLElement = any, ScrollContainer e
         rowsSelector,
     ]);
 
+    useLayoutEffectSafeForSsr(() => {
```
jakobz, line 60: Can we wrap all this with if (__DEV__), to throw away this some of this code in prod builds?

uui-core/src/hooks/useVirtualList/useVirtualList.ts
```diff
@@ -53,6 +57,20 @@ export function useVirtualList<List extends HTMLElement = any, ScrollContainer e
...
+        if (scrollContainer.current?.clientHeight
+            && prevScrollContainerClientHeight
+            && scrollContainer.current?.clientHeight !== prevScrollContainerClientHeight
+        ) {
+            ++scrollContainerHeightChangesCount.current;
+        }
+        if (scrollContainerHeightChangesCount.current > 20 && !scrollContainerHeightIsNotLimited.current) {
```
jakobz, line 67: Probably we can just trigger on some unrealistically tall viewport? E.g. if we see that we'll get 10k pixels tall - warn, and stop attempting to handle more?

PR #2536, opened 2024-10-07
added sending analytics by clicking on header cell

sanazez:


AlekseyManetov:
I believe we shouldn't add such amount of analytic event props to the parts of complex components like DataTable, let's try to track such events in DataSource api callbacks, here the example - https://codesandbox.io/p/sandbox/uui-forked-nctqyx

Issue #2544, opened 2024-10-11
[PickerModal] Modal stops working when "dataSource" prop is updated


siarheiyelin:
## Note
This functionality worked fine on  UUI 5.7.3 

## Steps to Reproduce
I've prepared a minimal example

```jsx
import { useContext, useState } from 'react';
import { PickerModal, FlexRow, FlexCell, Button, Checkbox } from '@epam/uui';
import { UuiContext, useLazyDataSource } from '@epam/uui-core';

type TItem = { id: string; name: string };

function api1() {
    const items = new Array(20)
        .fill(null)
        .map((_, index) => ({ id: `id_${index}`, name: `Name ${index}` }));
    return Promise.resolve({ items });
}

function api2() {
    const items = new Array(20).fill(null).map((_, index) => ({
        id: `id_${index + 2000}`,
        name: `Name ${index + 2000}`,
    }));
    return Promise.resolve({ items });
}

export function MainPage() {
    const context = useContext(UuiContext);

    const handleModalOpening = () => {
        context.uuiModals
            .show<TItem>((props) => {
                return <MyPickerModal {...props} />;
            })
            .then((newSelection) => {
                alert(`You selected: ${JSON.stringify(newSelection)}`);
            })
            .catch(() => {});
    };

    return (
        <FlexCell width={612}>
            <FlexRow columnGap="12">
                <Button
                    color="primary"
                    caption="Show"
                    onClick={handleModalOpening}
                />
            </FlexRow>
        </FlexCell>
    );
}

function MyPickerModal(props: any) {
    const [isUseAnotherApi, setIsUseAnotherApi] = useState<boolean>(false);

    const dataSource = useLazyDataSource<TItem, string, unknown>(
        {
            api: () => {
                if (isUseAnotherApi) {
                    return api2();
                }
                return api1();
            },
            getId: ({ id }) => id,
        },
        [isUseAnotherApi],
    );
    return (
        <PickerModal
            initialValue={undefined}
            dataSource={dataSource}
            selectionMode="single"
            valueType="entity"
            {...props}
            renderFilter={() => (
                <>
                    <Checkbox
                        value={isUseAnotherApi}
                        onValueChange={(newVal) =>
                            setIsUseAnotherApi(newVal === true)
                        }
                        label="Use another API"
                    />
                </>
            )}
        />
    );
}
```

Steps:
- Run the example using latest UUI lib
- Click button to open picker modal
- Check "Use another API" check box and try to select anything

## Actual result
[...]

PR #2546, opened 2024-10-11
[DataSources]: added refetching data on deps change.

Kuznietsov:
### Summary

Fixed https://github.com/epam/UUI/issues/2544


Kuznietsov:
@AlekseyManetov, maybe we should add an example with data source switching while using picker modals/inputs. WDYT?

PR #2649, opened 2024-11-20
[PickerInput]: fixed unnecessary api calls on body open with `minCharsToSearch` prop and search in body

AlekseyManetov:
#### QA notes: 

Need to retest PickerInput list loading behavior, initially selected items load and minCharsToSearch behavior in all cases(Pay attention, that it shouldn't be api calls in network when search length less than minCharsToSearch)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,10 @@
+# 5.xx.xx - xx.xx.2024
+
+**What's New**
+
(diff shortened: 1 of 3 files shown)


NatalliaAlieva:
Released in 5.12.0 ver.

Issue #1514, opened 2023-07-19
[MainMenu]: accessibility issues related to the component


NatalliaAlieva:
### Description
Menu items in the opened "More" menu are not accessible from the keyboard

### Precondition
Screen Reader is **OFF** or **ON**

### Steps to Reproduce
1. Open [Main Menu](https://uui.epam.com/documents?category=components&id=mainMenu&mode=doc&skin=UUI4_promo) component.
2. Use `Tab` key to navigate to collapsed "More" button and press `Enter` key to expand it.
3. Use `Tab` key to navigate to the opened items and observe.

### Actual Result
Menu items in the opened "More" menu are not accessible from the keyboard.

https://github.com/epam/UUI/assets/57751283/37d6cc36-a366-48bf-a8d3-975304b095e7

### Expected Result
All functionality including menu items in the opened "More" menu should be accessible from the keyboard.
**Failed WCAG Checkpoint:** 2.1.1 Keyboard (Level A)
https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html

### Description
Part of navigation links are not available when content is zoomed up to 200%

### Precondition
Screen Reader is **ON** or **OFF**

### Steps to Reproduce
1. Open [Main Menu](https://uui.epam.com/documents?category=components&id=mainMenu&mode=doc&skin=UUI4_promo) component.
2. Set the size of the adaptive panel so that the Main Menu (hamburger) button is displayed and click on it.
3. Zoom in the content to 200% and observe.

### Actual Result
"Action items" and "Subscriptions" links are not available when content is zoomed up to 200% and no vertical scroll is provided.

https://github.com/epam/UUI/assets/57751283/6cff37d7-b729-45d2-8128-7e1d2fe9b7f5

### Expected Result
All functionality should be available when content is zoomed up to 200%. 
**Failed WCAG Checkpoint:** 1.4.4 Resize text (Level AA)
https://www.w3.org/WAI/WCAG21/Understanding/resize-text.html

### Description
Incorrect implementation of aria attributes in navigation elements.

### Steps to Reproduce
1. Open [Main Menu](https://uui.epam.com/documents?id=mainMenu&mode=doc&category=components#responsive) component -> Responsive example.
2. Open DevTools and observe the code for the "More" navigation element.
3. Expand the "More" menu and observe the code for nested items.

### Actual Result
The role="menuitem" added to the for nested items (e.g. Companies). The aria-haspopup="true" and role=menuitem" added to the "More".
[...]

PR #2734, opened 2025-01-15
[MainMenu]: accessibility issues related to the component (case 6)

vasilii-kovalev:
### Description

* [MainMenuButton] Add `aria-current` attribute with value `page` for active links

#### Issue link

#1514, case 6

#### Additional resources

* [WAI-ARIA: `aria-current` attribute](https://www.w3.org/TR/wai-aria/#aria-current)
* [aria-current - Accessibility | MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current)


NatalliaAlieva:
Released in 5.13.0 ver.

PR #2767, opened 2025-02-26
Update RTEDemoApi.js

inkz:
This PR adds a `path.resolve` check to mitigate path traversal vulnerabilities. Previously, user-controlled input could potentially allow unauthorized file access by using `../` sequences.

#### QA notes:


AlekseyManetov:
@inkz Thanks for the fix!

Issue #2604, opened 2024-11-04
[PickerInput]: `Esc` key unexpectedly closes modal window when Picker Input dropdown is focused and expanded


NatalliaAlieva:
## Description
There is an issue within the modal window containing a picker input. When the dropdown of the single picker input is expanded and the ESC key is pressed, instead of just closing the dropdown, the entire modal window closes.

## Steps to Reproduce
1. Open [this](https://uui.epam.com/documents?id=modals&mode=doc&category=components#modal_with_form) example.
2. Expand "Country" by clicking on it.
3. Press the ESC key to close the picker.

## Actual result
When the `Esc` key is pressed while the dropdown of the picker input is open, the entire modal window closes along with the dropdown. 

https://github.com/user-attachments/assets/d0247f8f-6ef2-4030-b213-3fb20c85f2e5

## Expected result
Pressing the `Esc` key when the dropdown of the picker input is open should only close the dropdown, not the entire modal window. 

## Environment
* UUI version: 5.10.2
* Browser name: any
* Operating System: any

vasilii-kovalev:
## Solutions in other UI libraries

### [Radix UI](https://www.radix-ui.com)

[Similar issue in Radix UI](https://github.com/radix-ui/primitives/issues/1951).

Escapable components are wrapped in a dismissible layer with <kbd>Escape</kbd> button press handler. When <kbd>Escape</kbd> button is pressed, the layer's index is checked. If it is the highest one and the keyboard event has not been prevented already, call `event.preventDefault`. Otherwise, ignore the event.

[Implementation details](https://github.com/radix-ui/primitives/blob/main/packages/react/dismissable-layer/src/dismissable-layer.tsx#L102-L110).

### [React Aria](https://react-spectrum.adobe.com/react-aria/index.html)

`event.stopPropagation` and `event.preventDefault` call.

Also, there is a check for [`event.isComposing`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing).

[Composition event explanation](https://stackoverflow.com/a/51413620).

[Implementation details](https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/overlays/src/useOverlay.ts#L113-L119).

### [Material UI](https://mui.com/material-ui)

When <kbd>Escape</kbd> button is pressed, the layer's index is checked. If it is the highest one, call `event.stopPropagation`. Otherwise, ignore the event.

Close reason is passed alongside the event in the `onClose` callback.

[...]

PR #2839, opened 2025-06-02
[PickerInput]: `Esc` key unexpectedly closes modal window when Picker Input dropdown is focused and expanded

vasilii-kovalev:
### Description

After discussing the results of the [investigation](https://github.com/epam/UUI/issues/2604#issuecomment-2865279088) with @AlekseyManetov, it was decided to use `event.stopPropagation()` to prevent the event's bubbling up to the modal window.

The following changes were made:

* Stop "Escape" key press event propagation in Picker Input component when its body is opened
* Small code improvements in the touched files

#### Issue link
#2604

#### QA notes

### DIFF
--- a/uui/components/overlays/__tests__/Modals.test.tsx
+++ b/uui/components/overlays/__tests__/Modals.test.tsx
@@ -6,7 +6,7 @@ import { Button } from '../../buttons';
 import { Modals } from '@epam/uui-components';
 import { PickerInput } from '../../pickers';
 
-function TestElement(props?: ModalBlockerProps) {
+function TestElement(props: ModalBlockerProps) {
     return (
         <ModalBlocker { ...props }>
             <ModalWindow>
--- a/uui/components/overlays/__tests__/Modals.test.tsx
+++ b/uui/components/overlays/__tests__/Modals.test.tsx
@@ -1,9 +1,10 @@
-import React from 'react';
+import React, { ReactNode, useState } from 'react';
 import { ModalBlocker, ModalHeader, ModalFooter, ModalWindow } from '../Modals';
-import { fireEvent, renderSnapshotWithContextAsync, screen, setupComponentForTest } from '@epam/uui-test-utils';
-import { IModal, ModalBlockerProps, useUuiContext } from '@epam/uui-core';
+import { fireEvent, getDefaultUUiContextWrapper, renderSnapshotWithContextAsync, renderWithContextAsync, screen, setupComponentForTest } from '@epam/uui-test-utils';
+import { IModal, ModalBlockerProps, useArrayDataSource, useUuiContext } from '@epam/uui-core';
 import { Button } from '../../buttons';
 import { Modals } from '@epam/uui-components';
+import { PickerInput } from '../../pickers';
 
 function TestElement(props?: ModalBlockerProps) {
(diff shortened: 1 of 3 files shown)


uui/components/overlays/__tests__/Modals.test.tsx
```diff
@@ -1,11 +1,12 @@
...
-import { fireEvent, renderSnapshotWithContextAsync, screen, setupComponentForTest } from '@epam/uui-test-utils';
-import { IModal, ModalBlockerProps, useUuiContext } from '@epam/uui-core';
+import { fireEvent, getDefaultUUiContextWrapper, renderSnapshotWithContextAsync, renderWithContextAsync, screen, setupComponentForTest } from '@epam/uui-test-utils';
+import { IModal, ModalBlockerProps, useArrayDataSource, useUuiContext } from '@epam/uui-core';
 import { Button } from '../../buttons';
 import { Modals } from '@epam/uui-components';
+import { PickerInput } from '../../pickers';
 
-function TestElement(props?: ModalBlockerProps) {
+function TestElement(props: ModalBlockerProps) {
```
vasilii-kovalev, line 9: There was a TypeScript error, that `props` can be `undefined` when using `...props` or `props.abort`.
Judging by the component's usage (in `handleModalOpening` function below), the props can't be optional, so I removed the optional flag.

uui/components/overlays/__tests__/Modals.test.tsx
```diff
@@ -165,5 +166,155 @@ describe('Modals', () => {
...
+        ) {
+            interface Option {
+                id: string;
+                name: string;
+            }
+
+            const [
+                option,
+                setOption,
+            ] = useState<Option | undefined>(undefined);
```
vasilii-kovalev, line 185: Initially, I just passed `undefined` and an empty function to `value` and `onValueChange`, but decided to make the implementation as close as possible to how `PickerInput` is used in real world scenarios.

uui/components/overlays/__tests__/Modals.test.tsx
```diff
@@ -165,5 +166,155 @@ describe('Modals', () => {
...
+            ),
+        ).toBeVisible();
+
+        // First 'Escape' key press should close the picker input's options.
+        fireEvent.keyDown(
+            pickerInput,
+            {
+                key: 'Escape',
+            },
+        );
```
vasilii-kovalev, line 286: In other tests above, I can see that a new `KeyboardEvent` is created to fire an "Escape" key press event, though `fireEvent.keyDown` works fine.

uui/components/overlays/__tests__/Modals.test.tsx
```diff
@@ -165,5 +166,155 @@ describe('Modals', () => {
...
+        expect(
+            screen.queryByRole(
+                'option',
+                {
+                    name: /option 1/i,
+                },
+            ),
+        ).not.toBeInTheDocument();
+        expect(pickerInput).toBeInTheDocument();
+        expect(successMock).not.toBeCalled();
```
vasilii-kovalev, line 297: `success` is checked just to make sure, that this test will catch a regression, when it is called together with `abort` when closing the modal window (or even instead of it).

Issue #1553, opened 2023-08-01
[A11y]: component elements do not have programmatically determinable name for announcements 


NatalliaAlieva:
### Description
There is no announcement of role and state and changed state of Accordions

### Precondition
Screen Reader is **ON**

### Steps to Reproduce
1. Open [Accordion component](https://uui.epam.com/documents?id=accordion&mode=doc&skin=UUI4_promo&category=components).
2. Using `Tab` focus any of the presented Accordions examples.
3. Using `Enter`/`Space` key expand or collapse the section and note the announcement. 

### Actual Result
"[Name], group" is announced (e.g. Accordion block mode, group). There is no announcement of role and state. There is no announcement of changed state after pressing Enter/Space key. 

https://github.com/epam/UUI/assets/57751283/37b1b2f4-142d-4d11-9561-fdafc066ec0c

### Expected Result
Screen reader should announce accordion header name, role "button" and current state "collapsed" or "expanded" (e.g. Accordion block mode, collapsed, button). 
After pressing Space/Enter key screen reader should announce the changed state "collapsed" or "expanded".
Reference for implementation: https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/
**Failed WCAG Checkpoint:** 4.1.2 Name, Role, Value (Level A) 
https://www.w3.org/TR/WCAG21/#name-role-value

### Description
Plain text "Employee" on Custom accordion is announced as a button by screen reader

### Precondition
Screen Reader is **ON**

### Steps to Reproduce
1. Open [Accordion component](https://uui.epam.com/documents?id=accordion&mode=doc&skin=UUI4_promo&category=components).
2. Use the `down arrow` key to navigate to "Employee" text on "Custom accordion" and note the announcement. 

### Actual Result
Plain text "Employee" on "Custom accordion" is announced as a button by screen reader.
![image](https://github.com/epam/UUI/assets/57751283/26b3065d-a6c3-4a57-9f9c-0bada29ea53f)

### Expected Result
Plain text "Employee" on "Custom accordion" should not be announced as an interactive element button by screen reader.

**Failed WCAG Checkpoint:** 4.1.2 Name, Role, Value (Level A) 
https://www.w3.org/TR/WCAG21/#name-role-value

vasilii-kovalev:
After discussing the case 37 with UUI team, it was decided that it won't be fixed. Below you'll find an explanation why.

## Current solution

[...]

PR #2946, opened 2025-10-15
[A11y]: component elements do not have programmatically determinable name for announcements (case 12)

vasilii-kovalev:
### Description

* Provide text inputs inside data pickers with `aria-activedescendant` attribute

#### Issue link

#1553, case 12

#### Additional resources

* [Editable Combobox With List Autocomplete Example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/)

### DIFF
--- a/uui/components/pickers/DataPickerBody.tsx
+++ b/uui/components/pickers/DataPickerBody.tsx
@@ -74,22 +74,25 @@ export function DataPickerBody<TItem, TId>({ highlightSearchMatches = true, auto
         }
     };
 
+    const {
+        focusedIndex,
+        topIndex,
+    } = props.value;
+
     const focusedRowId = useMemo((): string => {
         // No need to make unnecessary calculations.
         if (!props.showSearch) {
             return '';
         }
 
-        const focusedRow = props.rows.find((row) => {
-            return row.isFocused;
-        });
+        const focusedRow = props.rows.at(focusedIndex - topIndex);
 
         if (!focusedRow) {
             return '';
         }
 
         return focusedRow.rowKey;
-    }, [props.showSearch, props.rows]);
+    }, [props.showSearch, focusedIndex, topIndex]);
 
     const renderEmpty = () => {
         const search = props.value.search;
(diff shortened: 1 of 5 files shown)


uui/components/pickers/DataPickerBody.tsx
```diff
@@ -49,6 +49,23 @@ export function DataPickerBody<TItem, TId>({ highlightSearchMatches = true, ...p
         if (e.shiftKey && e.key === 'Tab') e.preventDefault();
     };
 
+    const focusedRowId = useMemo((): string => {
+        // No need to make unnecessary calculations.
+        if (!props.showSearch) {
+            return '';
+        }
+
+        const focusedRow = props.rows.find((row) => {
```
vasilii-kovalev: Another approach of searching focused items would be to introduce `focusedId` (and change `id` value in `DataPickerRow`) or `focusedRowKey` alongside `focusedIndex` to avoid O(n) complexity. Is making the API more complex worth it in this case, @AlekseyManetov?

uui/components/pickers/PickerModal.tsx
```diff
@@ -98,6 +98,18 @@ export function PickerModal<TItem, TId>(props: PickerModalProps<TItem, TId>) {
 
     const dataRows = getRows();
 
+    const focusedRowId = useMemo((): string => {
```
vasilii-kovalev: Not sure how to reuse this logic.

AlekseyManetov (changes requested):

uui/components/pickers/DataPickerBody.tsx
```diff
@@ -49,6 +49,23 @@ export function DataPickerBody<TItem, TId>({ highlightSearchMatches = true, ...p
...
+            return row.isFocused;
+        });
+
+        if (!focusedRow) {
+            return '';
+        }
+
+        return focusedRow.rowKey;
+    }, [props.showSearch, props.rows]);
```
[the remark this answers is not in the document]
vasilii-kovalev: This approach doesn't work correctly - incorrect items are announced. It doesn't matter if `props.rows` is also present in the dependencies array or not. So the current implementation with `rows.find` is left as-is.

NatalliaAlieva:
Released in 6.4.0 ver.

Issue #1506, opened 2023-07-18
[PickerInput]: accessibility issues related to the component


NatalliaAlieva:
### Description
Mobile: Focus doesn't move to the opened dialog after Picker combobox opening.

### Precondition
Screen Reader (TalkBack on Android) is **ON**

### Steps to Reproduce
1. Open [Button](https://uui.epam.com/documents?id=button&mode=doc&skin=UUI3_loveship&category=components) component.
2. Use the swipe gesture to navigate to Picker combobox, double tap to expand it > "Person" dialog with options is opened, observe the focus placement.

### Actual Result
When Picker combobox is activated focus doesn't move to the opened dialog with options.

https://github.com/epam/UUI/assets/57751283/b7460451-9965-4c22-a928-d3a54c1e4aab

**NOTE:** reproducible using Bluetooth keyboard.

### Expected Result
When Picker combobox is activated, focus should be set to the "x" Close button in opened dialog with options. Screen reader should announce the role "dialog", name "Person" and currently focused Close button.
**Failed WCAG Checkpoint:** 2.4.3 Focus Order (Level A) 
https://www.w3.org/WAI/WCAG21/Understanding/focus-order.html

### Description
[CLEAR] button in single select combobox is not available using keyboard

### Precondition
Screen Reader is **OFF** or **ON**

### Steps to Reproduce
1. Proceed [Picker Input](https://uui.epam.com/documents?id=pickerInput&mode=doc&skin=UUI4_promo&category=components) component.
2. Use the `Tab` key to navigate to the Single select combobox and select any option.
3. Use the `Tab` key to navigate to [CLEAR] button.

### Actual Result
[CLEAR] button in expanded single select combobox is not available using keyboard.

https://github.com/epam/UUI/assets/57751283/35b436e5-6dc5-4305-80c0-3323adb7e4fe

**NOTE:** Reproducible for [SELECT ALL] button of "Persons" comboboxes in Search positions section, "Start type city name for search" and "Custom toggler" comboboxes in Picker toggler configuration and options section.

### Expected Result
[CLEAR] button in expanded single select combobox should be available using `Tab` key from the keyboard and should have a clear visible focus indicator.

**Failed WCAG Checkpoint:** 2.1.1 Keyboard (Level A)
https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html

### Description
When "Persons" combobox in Lazy list section is opened, pressing `Tab` key moves focus to the beginning of the page

### Precondition
[...]

PR #2947, opened 2025-10-15
[PickerInput]: accessibility issues related to the component (case 8)

vasilii-kovalev:
### Description

* Change value in `aria-haspopup` attribute from "true" to "dialog" in `PickerInput`

#### Issue link

#1506, case 8

#### QA notes


NatalliaAlieva:
Released in 6.4.0 ver.

PR #2948, opened 2025-10-15
[MainMenu]: accessibility issues related to the component (case 4)

vasilii-kovalev:
### Description

* Add `aria-haspopup="menu"` to the "More" button
* Add `role="menu"` to the container wrapping the list of items
* Provide an example in the documentation of how to assign `role="menuitem"` to the items only when they are present in the menu

#### Issue link

#1514, case 4

#### QA notes

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -16,7 +16,7 @@
 * [MainMenu] Improve accessibility of hidden items in "More" menu ([#2948](https://github.com/epam/UUI/pull/2948))
     * Add `aria-haspopup="menu"` to the "More" button
     * Add `role="menu"` to the container wrapping the list of items
-    * Provide an example in the documentation of how to assigning `role="menuitem"` to the items only when they are present in the menu
+    * Provide an example in the documentation of how to assign `role="menuitem"` to the items only when they are present in the menu
 
 # 6.3.1 - 08.10.2025
 


uui/components/navigation/MainMenu/MainMenuDropdown.tsx
```diff
@@ -37,6 +37,8 @@ export function MainMenuDropdown(props: MainMenuDropdownProps) {
                 </FocusLock>
             ) }
             placement="bottom-start"
+            // In the case, the default "dialog" role is not applicable.
+            role={ undefined }
```
vasilii-kovalev: Should we accept the role from props in this component instead of hard-coding the value, since the body can potentially be anything?

uui-components/src/overlays/Dropdown.tsx
```diff
@@ -440,7 +440,12 @@ function DropdownComponent(props: DropdownProps, ref: React.ForwardedRef<HTMLEle
                 <Portal target={ portalTarget }>
                     <FreeFocusInside>
                         <div
-                            role="dialog"
+                            role={
```
AlekseyManetov: why we need to remove  role="dialog" from main menu dropdown, but keep for all other dropdowns?

Issue #2945, opened 2025-10-13
[Dropdown]: add option to control dropdown positioning behavior


NatalliaAlieva:
## Now
Dropdowns are always repositioned (pinned) to the toggler, which may not be desirable in some cases.

https://github.com/user-attachments/assets/c750af11-95e3-4994-a999-99d82a1c7080

## To Do
Add a setting that allows the user to choose whether the dropdown should stay pinned to the toggler or remain static.

### Note
The default behavior needs to be discussed and defined separately.

PR #2960, opened 2025-11-13
[dropdown] added option to control dropdown positioning behavior

Margaryta-Maletz:
### Description:
[Dropdown]: add pinToToggler prop to control dropdown positioning relative to the toggler

#### Issue link:
https://github.com/epam/UUI/issues/2945

#### QA notes: 
[Dropdown]: added `pinToToggler` prop to control dropdown positioning. 
When set to `true`, the dropdown stays pinned to the toggler and follows its position. 
When set to `false`, the dropdown remains static at its initial position after opening. 
Default is `true`.

### DIFF
--- a/uui-components/src/overlays/Dropdown.tsx
+++ b/uui-components/src/overlays/Dropdown.tsx
@@ -33,13 +33,13 @@ function DropdownComponent(props: DropdownProps, ref: React.ForwardedRef<HTMLEle
         middleware,
         boundaryElement,
         closeOnEscape = true,
-        pinToToggler = true,
+        fixedBodyPosition = false,
     } = props;
 
     const uuiContext = useContext(UuiContext);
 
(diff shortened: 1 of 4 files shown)


uui-components/src/overlays/Dropdown.tsx
```diff
@@ -433,6 +435,20 @@ function DropdownComponent(props: DropdownProps, ref: React.ForwardedRef<HTMLEle
         updateTogglerSize();
     }
 
+    useEffect(() => {
+        if (!open) {
+            fixedPositionRef.current = null;
```
AlekseyManetov: I would suggest to rename fixedPositionRef to initialBodyPosition, to be more clear

Margaryta-Maletz: Done

PR #3007, opened 2025-12-29
[E2e tests]: DataTable check/uncheck functionality

Kuznietsov:
## Added tests for cascade selection functionality.


uui-e2e-tests/tests/Integration/dataTable/dataTable.select.e2e.ts
```diff
@@ -0,0 +1,147 @@
+import { expect } from '@playwright/test';
+import { test } from '../../../framework/fixtures';
+import { setupDocExampleTest } from '../testUtils';
+import { ColumnFiltersDataTableObject, LocationsDataTableObject } from '../../../framework/pageObjects';
+
+test.describe('DataTable: Select', () => {
```
AlekseyManetov: Let's also add tests for simple basic case with row selection in single and multi mode, without cascading(with and without keyboard)

PR #3094, opened 2026-05-07
feat(RangeDatePicker): create mobile view for RangeDatePicker

i-runets:
### Description:

Create a mobile version of the range date picker. The range date picker will automatically switch to the mobile version at 720px, which is the default value of the isMobile helper.

image

image

#### Issue link:

#3057 

#### QA notes:


JuliaMV (commented):
Nice improvement overall 🙌
Small parity note for mobile: onOpenChange isn’t fired when modal opens/closes, and modal Clear all currently ignores disableClear / preventEmptyFromDate / preventEmptyToDate (it always resets both dates to null).

MSt1ch (approved):

PR #3097, opened 2026-05-15
fix: keep notifications and recovery UI above deep modal stacks (#3095)

### LINKED ISSUE #3095: Notifications are displayed behind modal windows when there are a lot of layered components on the page
## Description

Some components create layers, like dropdowns, tooltips, or modal windows. When a layer is created, its z-index is calculated according to [this formula](https://github.com/epam/UUI/blob/f6da9778ecbb962b8197fa94c085537156a7529c/uui-core/src/services/LayoutContext.ts#L70) and then passed to the component that then uses it (in `style` property, for example).

The problem is, the notifications (and `Snackbar`) don't use that z-index and instead have a hardcoded value of 100500.
When there are a lot of components/elements on the page that create a layer (for example, a table with a lot of cells with dropdowns/tooltips in them) and the user opens a modal window, the modal window may have a z-index bigger than 100500, and the notifications that should be displayed as a result of actions within the modal window are displayed behind the modal window, which has a negative impact on the user experience.

Ideally, there should not be any hardcoded value, and the notifications (and/or `Snackbar`) should use the dynamic z-index from the layout context.

## Steps to Reproduce

I've created the following example:
* Cell count input manages the cell count. By default it is 1000. According to calculations (see the mentioned formula below), that issue should always be reproducible with a cell count >= 985 (though sometimes it is reproducible with lesser number)
* Table displays the grid of cells, whose number matches the cell count from the input. Each row has 25 columns to avoid cluttering
* Each cell contains a button with a cell number. On hover, a tooltip appears. On click, a modal window is opened
* The modal window contains a "Show notification" button. On click, a notification appears

1. Open [this CodeSandbox](https://codesandbox.io/p/devbox/uui-notification-layers-issue-mqkng5?file=%2Fsrc%2Fexample.tsx)
2. Hover over any cell in the table
3. Click on any cell
4. Click on the "Show notification" button in the modal window
5. Observe how the notification is displayed (in front of or behind the modal window blocker)
6. Close both modal window and notification, change the number of cells in the input to 300, and repeat steps 2-5

## Actual result

With 1000 cells, the notification is displayed behind the modal window.
With 300 cells, the notification is displayed in front of the modal window.

## Expected result

In both cases, the notification is displayed in front of the modal window.

JuliaMV:
## Summary

- Fixes [#3095](https://github.com/epam/UUI/issues/3095): with many stacked layers (e.g. large grids of tooltips/buttons), modal `z-index` could exceed the old fixed snackbar value (100500), so notifications appeared behind the modal.
- Snackbar / notification host stacking now follows `uuiLayout.getTopOverlayZIndex()`, keeping the previous value as a minimum so existing apps do not regress.
- `LayoutContext` exposes a stable top-overlay index based on registered layers; `ErrorHandler` recovery `ModalBlocker` uses the same source instead of a hardcoded `z-index`.

#### Issue link:
[3095](https://github.com/epam/UUI/issues/3095)

### DIFF
--- a/uui-components/src/overlays/Snackbar.tsx
+++ b/uui-components/src/overlays/Snackbar.tsx
@@ -11,6 +11,10 @@ const offset = 30;
 export interface SnackbarProps extends IHasCX, IHasRawProps<React.HTMLAttributes<HTMLDivElement>>, IHasForwardedRef<HTMLDivElement> {
     closeIcon?: Icon;
     notifications?: NotificationOperation[];
+    /**
+     * When set, applied as root `z-index` (e.g. from `uuiLayout.getTopOverlayZIndex()`).
+     * If omitted, styles use the default from CSS.
+     */
     zIndex?: number;
 }
 
--- a/uui-components/src/overlays/Snackbar.tsx
+++ b/uui-components/src/overlays/Snackbar.tsx
@@ -11,6 +11,7 @@ const offset = 30;
 export interface SnackbarProps extends IHasCX, IHasRawProps<React.HTMLAttributes<HTMLDivElement>>, IHasForwardedRef<HTMLDivElement> {
     closeIcon?: Icon;
     notifications?: NotificationOperation[];
+    zIndex?: number;
 }
 
 const uuiSnackbar = {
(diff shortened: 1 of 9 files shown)


uui-components/src/overlays/Snackbar.tsx
```diff
@@ -140,8 +141,22 @@ export class Snackbar extends React.Component<SnackbarProps> {
 
         const botCenterItems = items.filter((item: NotificationOperation) => item.config.position === 'bot-center').map(this.renderItemWithOffset(botCenterOffset));
 
+        const rawProps = this.props.rawProps ?? {};
+        const { style: rawStyle, ...restRawProps } = rawProps;
+        const rootStyle = this.props.zIndex !== undefined || rawStyle
+            ? {
+                ...(typeof rawStyle === 'object' && rawStyle ? (rawStyle as React.CSSProperties) : {}),
```
i-runets: I think we can skip this check for rawStyles type and just do this:

```
const rootStyle = rawStyle || this.props.zIndex !== undefined
    ? {
        ...(rawStyle ?? {}),
        ...(this.props.zIndex !== undefined ? { zIndex: this.props.zIndex } : {}),
    }
    : undefined;
```
rawStyle is already typed as React.CSSProperties(inherits from `IHasRawProps<React.HTMLAttributes<HTMLDivElement>>`), and from what I can see, UUI rarely does this kind of check.

JuliaMV: good point, thx!

i-runets (approved):

MSt1ch (approved):

Issue #3127, opened 2026-09-09
[TextPlaceholder]: component is empty on Safari


AHrakovich:
## Steps to Reproduce

just open component in Safari - https://uui.epam.com/documents?id=textPlaceholder&mode=doc&category=components&theme=promo

## Actual result

component is empty (no grey squares are displayed)

<img width="1448" height="1046" alt="Image" src="https://github.com/user-attachments/assets/898a3ccb-c35a-43ea-863b-fc4a8919dfc9" />

## Expected result

grey squares are displayed (depends on **wordCount** prop)

## Environment

* UUI version: 6.5.2
* Browser name: Safari (v 26.6.2)
* Operating System: macOS Tahoe 26.6.2

MSt1ch:
Released in v6.5.3

PR #3128, opened 2026-09-14
fix: TextPlaceholder is emty in Safari

i-runets:
### Description:

`TextPlaceholder` filled each fake "word" with repeated `&nbsp;` Safari suppresses ink for whitespace glyphs, so -webkit-background-clip: text had nothing to paint the shimmer gradient onto, making the bars invisible.

Fix: replace the &nbsp; filler with the letter x. In the Redacted font both characters share the same glyph/width, so layout is unchanged — but x isn't a whitespace character, so Safari renders it correctly.

#### Issue link:

#3127 

#### QA notes:


cpoftea (approved):

uui/components/typography/TextPlaceholder.tsx
```diff
@@ -32,8 +32,9 @@ export const TextPlaceholder: React.FunctionComponent<PropsWithChildren<ITextPla
                     className={ cx([
                         props.cx, css.loadingWord, !props.isNotAnimated && css.animatedLoading,
                     ]) }
-                    dangerouslySetInnerHTML={ { __html: it } }
```
JuliaMV, line 35: I would suggest to add aria-hidden="true" to avoid possible accessibility issues.
Additionality we should think about fallback incase the Redacted font fails to load.

i-runets: I agree on aria-hidden="true", but for the fallback we need to do a rework of the component — the shimmer relies on -webkit-background-clip: text, so it's coupled to whatever font/glyphs actually render, and swapping the fallback font stack alone won't fix that.

JuliaMV (approved):

